PR #1, opened 2017-04-24
Create WARNING_PATTERN regex to catch warnings

haypo:
Define a regular expression to catch different warnings, errors and
bugs specific to CPython.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -21,7 +21,8 @@ from buildbot.schedulers.forcesched import (ForceScheduler, FixedParameter,
                                             StringParameter, UserNameParameter)
 from buildbot.process import factory
 from buildbot.status import html, mail, words
-from buildbot.steps.shell import ShellCommand, Configure, Test, Compile
+from buildbot.steps.shell import (ShellCommand, Configure, Compile,
+                                  Test as BaseTest)
 from buildbot.steps.transfer import DirectoryUpload
 from buildbot import util, locks
 from buildbot.steps.source.git import Git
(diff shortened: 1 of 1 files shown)


haypo:
I didn't test my change.

I'm not sure about the last warning, PyErr_WriteUnraisable(), maybe we will catch too many warnings. Is it an issue? If yes, I can remove it. If not, we may remove it later if it's more noisy than useful.

haypo:
(oops, I clicked on the wrong button, I didn't want to close this PR.)

haypo:
cc @zware

haypo:
Documentation of the warningPattern parameter:
http://docs.buildbot.net/latest/manual/cfg-buildsteps.html#compile

haypo:
I want to get warnings from the "tests" step. I'm not sure that my change is correct.

zware:
I think `warnOnWarnings=True` on the `Test()` steps, with this regex, would cover that.  The default regex should be fine on the `Compile()` steps.

haypo:
Another interesting pattern: "Unhandled exception in thread" logged by the low-level t_bootstrap() function which manages the lifetime of a Python thread.

haypo:
> I think warnOnWarnings=True on the Test() steps, with this regex, would cover that. The default regex should be fine on the Compile() steps.

I checked buildbot code: Compile and Step classes have the same base class, and both support warningPattern. I updated my PR.

I created a new Test class to limit changes to avoid mistakes in the future in a new buildbot is added.

haypo:
> warnOnWarnings=True

I don't want to make the buildbot fail on these new warnings. I will first analyze them and try to fix some of them. It's ok to get *warnings* ;-)

zware:
> I don't want to make the buildbot fail on these new warnings.

Right, there's `failOnWarnings` for that :).  `warnOnWarnings` should just make it orange rather than green, but still a pass.

haypo:
> Looks fine to me, we'll give it a try.

Oh, I'm suprised: it seems like my change worked :-) Did you have to run a chef/puppet job for that? Or reload buildbot configuration manually? Or everything is automated?

Example of catched "Fatal Python error: Aborted":

http://buildbot.python.org/all/builders/AMD64%20OpenIndiana%203.x/builds/467/steps/test/logs/warnings%20%2817%29

Example of catched "Warning -- files was modified by test_distutils":

http://buildbot.python.org/all/builders/AMD64%20Windows10%203.x/builds/595/steps/test/logs/warnings%20%281%29

These warnings are very valuable to me!

PR #2, opened 2017-04-26
warningPattern: add Windows fatal exception

haypo:
* Fix also the regex for ResourceWarning: need ".*" prefix, since
  .match() is used, not .search()
* Compile warningPattern regex once

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -11,6 +11,7 @@
 # buildmaster. They are documented in docs/config.xhtml .
 
 import os
+import re
 import sys
 
 from twisted.python import log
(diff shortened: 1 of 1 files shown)


haypo:
My first patch #1 worked well! But ResourceWarning warnings are not catched because of a bug in my regex, and I also noticed another interesting warning:

http://buildbot.python.org/all/builders/AMD64%20Windows8.1%20Non-Debug%203.x/builds/670/steps/test/logs/stdio

```
0:06:33 [299/404/2] test_multiprocessing_fork crashed (Exit code 3221225477) -- running: test_tools (65 sec)
Windows fatal exception: access violation

Current thread 0x00000850 (most recent call first):
  File "D:\buildarea\3.x.ware-win81-release\build\lib\test\_test_multiprocessing.py", line 3992 in ManagerMixin
  File "D:\buildarea\3.x.ware-win81-release\build\lib\test\_test_multiprocessing.py", line 3988 in <module>
...
```

@zware: I'm interested to be allowed to push directly if you explain me steps required to apply the new config on buildbot, or we can continue with pull requests. It's up to you!

PR #3, opened 2017-05-06
Add Travis CI

zware:


haypo:
Do you suggest me to now use pull requests for conf changes?

zware:
I did this in a PR just so I could test it and easily throw it away if it didn't work; I'm not too bothered whether you use PRs for everything or not as long as you don't break it :)

PR #4, opened 2017-05-09
bpo-30314: Give bolen-tiger extra test time

zware:
@haypo Does this look like a reasonable resolution to [bpo-30314](https://bugs.python.org/issue30314)?

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -600,6 +600,10 @@ extra_factory_args = {
         # test curses as well
         'test_with_PTY': True,
     },
+    'bolen-tiger': {
+        # test_tools is very slow sometimes, give it extra time
+        'test_timeout': 30 * 60,
+    },
 }
 
 # The following with the slave owners' agreement


haypo:
I dislike this change. I would prefer to increase the default timeout for all buildbots, not the faulthandler (test) timeout (15 min), but also the build timeout (1h?). For example 15 min => 30 min, 1h => 1h30?

zware:
Adjusting the timeout value this way changes both the buildbot timeout and the faulthandler timeout: the faulthandler timeout defaults to 5 minutes less than the buildbot timeout, see [lines 87-90](https://github.com/python/buildmaster-config/blob/master/master/master.cfg#L87-L90), [191-204](https://github.com/python/buildmaster-config/blob/master/master/master.cfg#L191-L204), and [372-379](https://github.com/python/buildmaster-config/blob/master/master/master.cfg#L372-L379).

I'm ok with bumping up timeouts across the board (that would be line 90), but it is nice to have a fairly quick timeout for cases where something is truly broken and hanging.

haypo:
> but it is nice to have a fairly quick timeout for cases where something is truly broken and hanging.

Hanging became much less common nowadays, Python became much more
stable since 2 years. Last months, I see more likely builds killed
because the buildbot was highly loaded and so some tests took longer
than 15 min.

See also http://bugs.python.org/issue30172 discussion. See also the
"10 slowest tests" output on buildbots.

PR #5, opened 2017-05-10
Send notifications to buildbot-status@python.org

haypo:


haypo:
buildbot@python.org is not subscribed to https://mail.python.org/mm3/mailman3/lists/buildbot-status.python.org/

Right now, mails sent from non-member of the list are hold for moderation. We should fix this issue first.

haypo:
> Right now, mails sent from non-member of the list are hold for moderation. We should fix this issue first.

I subscribed buildbot@python.org manually to the list, but I'm unable to disable delivery to this address :-( I don't know if buildbot@python.org has an inbox?

PR #8, opened 2017-05-20
Adjust builder timeouts

haypo:
* UnixRefleakBuild: use timeout x 10 since --huntrleaks makes tests
  much slower
* bpo-30314: Use 30 min rather than 20 min for x86 Tiger.


haypo:
I just applied the new conf. I will try to check if the config was applied correctly.

PR #9, opened 2017-05-20
Disable the `cpu` resource on refleak builds

zware:


zware:
Travis has been sitting on this for 41 minutes now; I'm impatient so I disabled the requirement.

PR #16, opened 2017-07-26
Remove koobs-freebsd9 (9-STABLE) builder

koobs:
The releng/9.3 (latest 9.x release security branch) and stable/9 branches (the latter of which this worker run on), hit End-of-Life December 31, 2016 [1]

Accordingly, this change removes the koobs-freebsd9 builder as support for it has stopped.

[1] https://www.freebsd.org/security/unsupported.html

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -462,7 +462,6 @@ if PRODUCTION:
         # OS X
         ("x86 Tiger", "bolen-tiger", TigerBuild, STABLE),
         # Other Unix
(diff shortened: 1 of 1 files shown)


haypo:
Yeah, you already told me. I'm not sad, we had too many bugs really specific to FreeBSD 9, so for me, it will reduce the number of buildbot bugs to handler :-)

Hopefully, FreeBSD keeps me busy enough: a *new* random bug everyday! Today one is makedev:
http://bugs.python.org/issue31044

Good bye FreeBSD 9.

PR #17, opened 2017-08-08
Remove Python 3.5 buildbots

haypo:
Python 3.5 just entered security-fixes only:
https://mail.python.org/pipermail/python-dev/2017-August/148794.html


haypo:
Goodbye Python 3.5 buildbot slaves! It will reduce the pressure on buildbots!

PR #18, opened 2017-08-11
Drop OpenIndiana slaves

haypo:
The builder is offline since the beginning of June.

Previous discussions on python-dev:

* September 2016: OpenIndiana and Solaris support
  https://mail.python.org/pipermail/python-dev/2016-September/146538.html
* April 2015: MemoryError and other bugs on AMD64 OpenIndiana 3.x
  https://mail.python.org/pipermail/python-dev/2015-April/138967.html
* September 2014: Sad status of Python 3.x buildbots
  https://mail.python.org/pipermail/python-dev/2014-September/136175.html

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -254,18 +254,6 @@ class SharedUnixBuild(UnixBuild):
     configureFlags = ["--with-pydebug", "--enable-shared"]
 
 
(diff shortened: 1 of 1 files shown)


haypo:
cc @zware and @jcea.

PR #33, opened 2018-05-09
Add a Clang UBSan build slave.

gpshead:
Adds a Clang Undefined behavior sanitizer.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -126,13 +126,15 @@ class FreezeBuild(TaggedBuildFactory):
 
 class UnixBuild(TaggedBuildFactory):
     configureFlags = ["--with-pydebug"]
+    configureEnv = {}
     interpreterFlags = ""
     testFlags = "-j2"
     makeTarget = "all"
     test_timeout = None
 
     def setup(self, parallel, test_with_PTY=False, **kwargs):
-        self.addStep(Configure(command=['./configure'] + self.configureFlags))
+        self.addStep(Configure(command=['./configure'] + self.configureFlags,
+                               env=self.configureEnv))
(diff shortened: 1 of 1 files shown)


gpshead:
Let me know if there is more I need to do?

master/master.cfg
```diff
@@ -126,13 +126,15 @@ class FreezeBuild(TaggedBuildFactory):
 
 class UnixBuild(TaggedBuildFactory):
     configureFlags = ["--with-pydebug"]
+    configureEnv = {}
```
vstinner: It seems like an empty dict is fine, since it's only *added* environment variables, it doesn't *replace* the full environment as env= of subprocess.Popen.

master/master.cfg
```diff
@@ -249,6 +251,15 @@ class PGOUnixBuild(NonDebugUnixBuild):
     factory_tags = ['pgo']
 
 
+class ClangUBSanLinuxBuild(UnixBuild):
+    configureEnv = UnixBuild.configureEnv.copy()
```
vstinner: You might use dict(**UnixBuild.configureEnv, CC='clang', ...) but I'm not sure it's more readable.

zware: Alternately, these could be passed in `configureFlags` (`'CC=clang', 'LD=clang', ...`), just to not add `configureEnv`.  I'm fine with giving this a shot as is, though.

gpshead: I was just replicating what I did on the command line that had worked.  if they work as parameters, I'll do that, I'll double check here and simplify to just use the params if so...

gpshead:
PTAL.  should be good now.

PR #35, opened 2018-05-23
Try to fix syntax of the SlowWindowsBuild options.

ned-deily:


vstinner:
I applied manually (as usual) the new configuration.

ned-deily:
It looks the configuration hasn't been updated yet.  @zware, could you take a look at it, please?

ned-deily:
I believe this is fixed now, thanks @zware.

PR #47, opened 2018-09-09
Modify AIX bots to use less optimized OpenSSL library

aixtools:
as the highly optimized openssl supplied by IBM does not always
report the error message, only that an error has occurred.

After I built this library from OpenSSL sources and manually tested python for AIX linked against it all the ssl.py and _ssl.c dependent tests pass.

While this is not the library that I expect to be used in production it does demonstrate that AIX is stable with regard to OpenSSL (tests).

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -232,13 +232,17 @@ class UnixBuildWithoutDocStrings(UnixBuild):
 class AIXBuild(UnixBuild):
     configureFlags = [
         "--with-pydebug",
-        "--without-gcc",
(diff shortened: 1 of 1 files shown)


zware:
I'm happy to accept this for your worker, but I'd like to hear from @edelsohn before modifying his worker.

edelsohn:
It's fine with me to use the corrected library for my worker (assuming that it functions correctly).

zware:
Good enough for me :)

PR #50, opened 2018-09-13
Add GitHub status check config

zware:


pablogsal:
One question: If a buildbot is building multiple commits all of them will be reported in GitHub?

zware:
That is an excellent question that I will ask on #buildbot on freenode, but I don't expect to know until it happens :)

PR #51, opened 2018-09-14
Add all stable builders to GitHubStatus reporter

zware:


pablogsal:
I wonder if an extra label (like `GITHUB BUILDER`) could make sense in case we want to separate the stable builders from the GitHub ones in the future.

zware:
Adding a `github` tag is a really good idea, actually.  I'll implement that and push it here.

zware:
Actually, digging into I haven't hit upon a clean way to add the tag that I like without rewriting most of the file, so for now I think I'll stick with this.  That is a suggestion I'll keep in mind the next time this gets touched, though.

PR #57, opened 2018-09-30
Add Pull Request GitHub reporter

pablogsal:
This Pull Request adds a GitHub reporter that, in case of a build failure, will issue a comment in the Pull Request of the original commit to raise awareness.

To reduce the noise, the following conditions must be satisfied:

- The build is a falure
- The build has sourcestamps
- The build does not have more than 1 change. (This will avoid the problem of builds that build multiple commits at the same time).
- We can detect the Pull Request from the commit message. We extract the PR number
using the regular expression `\((?:GH-|#)(\d+)\)`. In case of multiple matches we use the last one added (to account for backports).

The message is probably not the best one but I did not have a very creative night.  Feel free to suggest improvements.

You can see an example of the report here:

https://github.com/pablogsal/cpython/pull/2

### DIFF
--- a/master/custom/pr_reporter.py
+++ b/master/custom/pr_reporter.py
@@ -1,6 +1,3 @@
-from __future__ import absolute_import
-from __future__ import print_function
-
 import textwrap
 import re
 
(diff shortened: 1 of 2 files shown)


pablogsal:
I will fix this failure today.

pablogsal:
@zware Thanks for the review. I have addressed the feedback in f535bda.

pablogsal:
Btw, it **may** be possible that this needs `txrequests` as it can use two different backends, being `txrequest` the most reliable in my experiments.

pablogsal:
@zware Gentle ping

PR #67, opened 2018-11-28
Only notify failures by email for stable builders

pablogsal:
Some builders are sending email notifications but they are fundamentally broken. An example of this is the AMD64 Cygwin64 on Windows 10 3.x:

https://buildbot.python.org/all/#/builders/164

As we already have a lot of noise due to random failures, minimizing the ammount of email we need to check is important.


pablogsal:
CC @vstinner

PR #75, opened 2019-02-10
Rename ARMv7 Ubuntu buildbot to ARMv7 Debian

gpshead:
Because I upgraded the ancient OS.

The `gps-ubuntu-exynos5-armv7l` name is also laughably wrong as that was the original long gone hardware I setup in early 2013 but most people never see that name (it isn't the title on the buildbot pages) so i'm not worried about changing it for now.


gpshead:
Thanks! I don't have merge permissions in this repo, someone please merge (or grant me permission).

pablogsal:
I will deploy this changes tomorrow morning. :)

(I do not have permission to give privileges in GitHub so I just merged the PR)

zware:
I currently have merge permissions restricted to those with the ability to deploy the changes to avoid having a bunch of changes build up before someone gets around to deploying them.  If (when?) we get things better set up such that deployment is automatic, we'll likely grant wider merge permissions.

PR #79, opened 2019-03-04
Fix typo for environment variable  from $(PWD) to ${PWD}

aixtools:
Hope this helps!

p.s. checked that bash also tries to execute PWD with $(PWD) rather than magically provide the expected result.


zware:
`$(PWD)` is currently working; I believe it is expanded by the Makefile rather than by Bash.  Since what we have is working as expected, I'm going to reject this PR.

PR #81, opened 2019-03-04
syncronize factory classname schemes for AIXBuild with other platforms

aixtools:
The classnames are incorrect - imho. The new names are more inline other platforms (and an incorrect argument that was the basis for one classname is removed).

### DIFF
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -26,7 +26,7 @@ class FreezeBuild(TaggedBuildFactory):
     factory_tags = ["freeze"]
 
     def setup(self, **kwargs):
-        self.addStep(Configure(command=["./configure", "--prefix", "${PWD}/target"]))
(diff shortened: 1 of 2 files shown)


master/custom/factories.py
```diff
@@ -125,7 +125,7 @@ def setup(self, parallel, branch, test_with_PTY=False, **kwargs):
         installed_python = "./target/bin/python%s" % branch
         self.addStep(
             Configure(
-                command=["./configure", "--prefix", "$(PWD)/target"]
+                command=["./configure", "--prefix", "${PWD}/target"]
```
zware: Please revert the `${PWD}` changes; `$(PWD)` is currently working as expected for us.  IIRC, I tried `${PWD}` initially and did not get the expected result.  See also my comment on closing #79.

PR #86, opened 2019-04-10
Disable notifiations on broken workers

vstinner:
Disable email and IRC notifications on workers which were never green
(tests always failed).


vstinner:
cc @pablogsal

vstinner:
New config deployed.

PR #87, opened 2019-04-11
Add the reference leak build and the clang build to the Fedora Rawhide worker.

stratakis:


stratakis:
Closing this in favor of #88

PR #88, opened 2019-04-12
Add ClangUnixInstalledBuild. Also add the clang and refleak builds for the AMD64 Fedora Rawhide worker

stratakis:
Add a build config for running tests with
an installed python compiled with clang

### DIFF
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -209,6 +209,12 @@ class ClangUbsanLinuxBuild(UnixBuild):
     testFlags = "-j4"
     factory_tags = ["clang", "ubsan", "sanitizer"]
 
+class ClangUnixInstalledBuild(UnixInstalledBuild):
+    configureFlags = [
+        "CC=clang",
+        "LD=clang",
+    ]
+    factory_tags = ["clang", "installed"]
 
 class SharedUnixBuild(UnixBuild):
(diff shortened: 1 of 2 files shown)


vstinner:
I deployed the new config with an additional minor fix:

```
commit fd08bbf525d55b10c2eba7916534eb20e06a4232
Author: Victor Stinner <vstinner@redhat.com>
Date:   Fri Apr 12 14:50:04 2019 +0200

    Add missing import

diff --git a/master/master.cfg b/master/master.cfg
index ca10108..eed0070 100644
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -36,6 +36,7 @@ from custom.factories import (
     NonDebugUnixBuild,
     PGOUnixBuild,
     ClangUbsanLinuxBuild,
+    ClangUnixInstalledBuild,
     SharedUnixBuild,
     WindowsBuild,
     SlowWindowsBuild,
```

I'm not sure how Travis CI missed this bug?

zware:
Unfortunately, Travis doesn't check the "production" config since it doesn't have configuration for all of the workers.  It's on my "I need to do this eventually" list to move the worker config from `local.py` to `master.cfg` with just the secrets stored in `local.py`, but it hasn't happened yet.

vstinner:
Ok. Don't worry, it took me 30 seconds to fix the obvious issue ;-)

PR #89, opened 2019-04-23
assign Windows arm32 worker

paulmon:
I expect this worker to fail at first.
For example tools/buildbot/test.bat needs to changed to support this worker: https://github.com/python/cpython/pull/12917
There are also 10 more tests that need skips or fixes that I'm aware of.

### DIFF
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -369,3 +369,17 @@ class Windows64ICCReleaseBuild(Windows64ReleaseBuild):
     buildersuffix = ".nondebug"
     buildFlags = Windows64ReleaseBuild.buildFlags + windows_icc_build_flags
     factory_tags = ["win64", "icc", "nondebug"]
+
+class WindowsArm32Build(WindowsBuild):
+    buildFlags = ["-p", "ARM", "--no-tkinter", "--no-ssl"]
+    # test_multiprocessing_spawn doesn't complete over simple ssh connection
+    # skip test_multiprocessing_spawn for now
+    testFlags = ["-arm32", "-j2", "-x", "test_multiprocessing_spawn"]
+    cleanFlags = ["-p", "ARM", "--no-tkinter", "--no-ssl"]
+    factory_tags = ["arm32"]
+
+class WindowsArm32ReleaseBuild(WindowsArm32Build):
+    buildFlags = WindowsArm32Build.buildFlags + ["-c", "Release"]
+    testFlags = WindowsArm32Build.testFlags + ["+d"]
+    # keep default cleanFlags, both configurations get cleaned
+    factory_tags = ["arm32", "nondebug"]
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -374,7 +374,7 @@ class WindowsArm32Build(WindowsBuild):
     buildFlags = ["-p", "ARM", "--no-tkinter"]
     # test_multiprocessing_spawn doesn't complete over simple ssh connection
     # skip test_multiprocessing_spawn for now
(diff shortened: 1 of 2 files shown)


master/custom/factories.py
```diff
@@ -369,3 +369,17 @@ class Windows64ICCReleaseBuild(Windows64ReleaseBuild):
     buildersuffix = ".nondebug"
     buildFlags = Windows64ReleaseBuild.buildFlags + windows_icc_build_flags
     factory_tags = ["win64", "icc", "nondebug"]
+
+class WindowsArm32Build(WindowsBuild):
+    buildFlags = ["-p", "ARM", "--no-tkinter", "--no-ssl"]
```
zooba: Didn't we get SSL working?

paulmon: @zooba 
Yes.  But it's not merged yet.
The PR is here: https://github.com/python/cpython/pull/11779

zooba:
Okay, I'm making a new OpenSSL build with that fix.

If the only problem with the buildbot is that the tests fail, I'd rather go in with the failure. It's marked unstable for now, so it's okay, and it'll both save updating the config when it's fixed and will force us to make the fixes :)

paulmon:
@zooba, Are you saying to remove the --no-ssl parameter now and let it fail until the ssl PR is merged?

zooba:
Yep, exactly.

zooba:
Ping @zware @pablogsal @vstinner - look like you all have commit rights?

We're about to get the SSL changes in (I'm running them through the current buildbot fleet), and at that point we have no reason not to include Paul's machines in the unstable fleet.

master/custom/factories.py
```diff
@@ -369,3 +369,17 @@ class Windows64ICCReleaseBuild(Windows64ReleaseBuild):
...
     buildFlags = Windows64ReleaseBuild.buildFlags + windows_icc_build_flags
     factory_tags = ["win64", "icc", "nondebug"]
+
+class WindowsArm32Build(WindowsBuild):
+    buildFlags = ["-p", "ARM", "--no-tkinter"]
+    # test_multiprocessing_spawn doesn't complete over simple ssh connection
+    # skip test_multiprocessing_spawn for now
+    testFlags = ["-arm32", "-j2", "-x", "test_multiprocessing_spawn", "-x", "test_winconsoleio", "-x", "test_distutils"]
+    cleanFlags = ["-p", "ARM", "--no-tkinter"]
+    factory_tags = ["arm32"]
```
zware: We need some manner of Windows tag here too.  Maybe `win-arm32` instead of `arm32`, or a separate `windows` tag?  Same below.

zooba:
@zware Changes have been made - anything else?

master/master.cfg
```diff
@@ -218,6 +220,8 @@ if PRODUCTION:
         ("POWER6 AIX", "aixtools-aix-power6", AIXBuild, UNSTABLE),
         ("PPC64 AIX", "edelsohn-aix-ppc64", AIXBuildWithGcc, UNSTABLE),
         # Windows
+        ("ARM32 Windows10 1903", "monson-win-arm32", WindowsArm32Build, UNSTABLE),
+        ("ARM32 Windows10 1903", "monson-win-arm32", WindowsArm32ReleaseBuild, UNSTABLE),
```
zware:
```suggestion
        ("ARM32 Windows10 1903 Non-Debug", "monson-win-arm32", WindowsArm32ReleaseBuild, UNSTABLE),
```

paulmon:
@zware Thanks for reviewing.  I merged your suggested change

PR #91, opened 2019-05-20
bpo-36511: Fix Windows arm32 buildbot, pythoninfo, scp, and ssh

paulmon:
When running the buildbot for Windows ARM32 test.pythoninfo must be run remotely.
Also seperating deploy step from both pythoninfo and test.bat
This requires a cpython PR to work: https://github.com/python/cpython/pull/13454
I think this should work.  Feedback is appreciated.

@zooba @zware

### DIFF
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -260,14 +260,19 @@ class DTraceCLangBuild(UnixBuild):
 
 class WindowsBuild(TaggedBuildFactory):
     build_command = [r"Tools\buildbot\build.bat"]
+    remote_deploy_command = [r"Tools\buildbot\remoteDeploy.bat"]
+    remote_pythonInfo_command = [r"Tools\buildbot\remotePythoninfo.bat"]
     test_command = [r"Tools\buildbot\test.bat"]
     clean_command = [r"Tools\buildbot\clean.bat"]
     python_command = [r"python.bat"]
     buildFlags = []
+    remoteDeployFlags = []
+    remotePythonInfoFlags = []
     testFlags = ["-j2"]
     cleanFlags = []
     test_timeout = None
     factory_tags = ["win32"]
+    remoteTest = False
 
     def setup(self, parallel, branch, **kwargs):
         build_command = self.build_command + self.buildFlags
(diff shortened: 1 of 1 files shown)


zooba:
I think this could also be done by handling the flag specially in the test.bat file, yes? Rather than making more modifications here.

I'll leave it to @zware as to where he'd like the split to be - I don't have a strong preference.

paulmon:
I found that test.pythoninfo is called directly from a buildbot step without calling any of the .bat files in tools\buildbot currently.  This is why I didn't catch it before.

master/custom/factories.py
```diff
@@ -276,14 +281,34 @@ def setup(self, parallel, branch, **kwargs):
...
+                    command=self.remote_pythonInfo_command + self.remotePythonInfoFlags,
+                    warnOnFailure=True,
+                )
+            )
+        else:
+            self.addStep(
+                ShellCommand(
+                    name="pythoninfo",
+                    description="pythoninfo",
+                    command=self.python_command + ["-m", "test.pythoninfo"],
```
paulmon: This line doesn't call any .bat files.  It ends up executing an arm32 binary on the x86/x64 build host which doesn't work

paulmon:
The must-have parts of this PR are that the test files must be copied to the test device before **pythoninfo** and **pythoninfo** must be run on the test device, not the build host.  Currently it's not possible to change where **pythoninfo** runs without changing the buildmaster-config

paulmon:
In my manual testing it looks to me like this change with these 2 changes will get the arm32 buildbot to zero test failures
https://github.com/python/cpython/pull/13454 
https://github.com/python/cpython/pull/13073

I will be away until Mon Jun 3, but I'd like to get this wrapped up soon if possible.

Thanks,
Paul

@zware @zooba

zooba:
@zware Need you to approve this, but it looks okay to me.

zware:
Sorry for the delay.  This was not the solution I'd have preferred for this issue, but I'm not going to have time to work that out and it's not fair to try to make you do it when I'm not even sure it would work :).  This will work and will solve the issue, so we'll go with it.

paulmon:
Thanks! If you have suggestions for making it better I am interested in helping out.

zware:
What I think I'd like to see is separate builders for the 'build' and 'test' portions running on separate workers with the 'test' builders triggered by success on the 'build' builders, so that everything can just be native commands without dealing with SSH or special batch files.  This does require `buildbot-worker` to actually run on an ARM host which could be a bit of a challenge, and I'm not sure how to transfer the built files from one worker to another (possibly packaged up installer-like, and also made available via the buildbot master's web server?), but it seems rather cleaner to me :)

paulmon:
That sounds like a great idea.  I can't get to right now, but I'll keep it mind

PR #98, opened 2019-06-24
Add CleanupTest step

vstinner:


vstinner:
Implement https://bugs.python.org/issue37359

vstinner:
I deployed the new config.

PR #103, opened 2019-07-29
[ON HOLD] Change the fedora rawhide worker naming to be more specific

stratakis:
Pending merge till the non-public bits of the infrastructure get renamed as well.


vstinner:
I already renamed cstratak-fedora to cstratak-fedora-rawhide-x86_64: https://github.com/python/buildmaster-config/commit/b1a3ee3e93a8191557fdb236125bfb52f53ddb63

PR #108, opened 2019-09-17
Fix build config suffixes

stratakis:
Build configs that inherit from the same configuration need a suffix in order
to not duplicate the build directory

PR #109, opened 2019-09-17
Remove old unused factories

vstinner:


vstinner:
While reviewing PR #108, I noticed that many factory classes are no longer used. I propose this change to remove them.

PR #111, opened 2019-09-25
Fix a typo in the clang ubsan buildersuffix.

gpshead:
This will make the disk space problems caused by the earlier change worse, but this already broke this buildbot so rather than keep a typo in the name we may as well make it correct.  https://bugs.python.org/issue38269


vstinner:
I merged your PR and deployed the new config, thanks for this fix.

PR #112, opened 2019-09-25
Fix a typo in ClangUbsanLinuxBuild suffix

vstinner:
It should be read as "UB san" (Undefined Behavior Sanitizer)


vstinner:
Oh, I didn't notice that @gpshead already wrote a similar PR. So I merged his PR #111.

PR #115, opened 2019-10-20
Rename freebsd workers and builders

zware:


zware:
@koobs, can you make sure I got this straight?

koobs:
@zware Looks fine

PR #116, opened 2019-10-22
Switch from secret local.py file to secret config file

zware:
The config file can be rather more minimal, and should be easier to generate via Salt.  I'm not trilled with my implementation of the `Settings` class, but it basically does what I want it to do.  I considered using [buildbot's own](https://docs.buildbot.net/2.5.0/manual/secretsmanagement.html) secret management, but none of this is passed to the workers, I'm reluctant to set up a new secret storing service, and I think it would be harder to manage one-file-per-secret rather than one file with all of the secrets.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ buildbot.sh
 buildbot/
 lib/
 venv/
+worker/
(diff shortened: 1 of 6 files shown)


vstinner:
I don't know Salt. I'm not available to review this change. If it works, well, just merge it ;-)

PR #121, opened 2019-11-28
Clean up daily builder scheduling

zware:
The main goal here is to split up the long refleak builds throughout the day, so other builds on the refleak hosts hopefully aren't as blocked by refleak builds.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -422,7 +422,7 @@ for branch_num, (git_url, branchname, git_branch) in enumerate(git_branches):
     if dailybuildernames:
         c['schedulers'].append(Nightly(
             name=branchname + '-daily',
-            hour=int(24 / (len(git_branches) - 1) * branch_num),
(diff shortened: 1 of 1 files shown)


master/master.cfg
```diff
@@ -419,13 +419,14 @@ for git_url, branchname, git_branch in git_branches:
...
-        name=branchname + '-daily',
-        hour=0,
-        minute=0,
-        change_filter=util.ChangeFilter(branch=git_branch),
-        builderNames=dailybuildernames,
-    ))
+    if dailybuildernames:
+        c['schedulers'].append(Nightly(
+            name=branchname + '-daily',
+            hour=int(24 / (len(git_branches) - 1) * branch_num),
```
vstinner: Hum, do we get hour=24 for the last Git branch? Should we limit to hour=23?

```suggestion
            hour=int(branch_num / (len(git_branches) - 1) * 23),
```

PR #124, opened 2019-12-17
Don't pass --timeout on 2.7.

benjaminp:
2.7's regrtest doesn't support --timeout.


zware:
Ah, whoops.  Thanks for catching that.

PR #125, opened 2019-12-28
Add configuration for an Arch Linux worker

pablogsal:


master/custom/workers.py
```diff
@@ -147,6 +147,11 @@ def get_workers(settings):
             tags=['windows', 'win10', 'arm', 'arm32'],
             branches=['3.x'],
         ),
+        cpw(
+            name="pablogsal-arch-x86_64",
+            tags=['linux', 'unix', 'arch', 'x86-64'],
```
zware: Is this 32-bit or 64-bit?  If 32, the tag should be `x86`; if 64, there should also be `amd64`.

Note that I don't think we're actually using these tags anywhere yet, so it doesn't *really* matter :)

It does have implications for the builder name, though; the `x86` in that name makes it sound like 32-bit, but the `x86_64` in the worker name makes it sound like 64 bit.

pablogsal: Is 64-bit (intel processor). Thanks for the heads up, I will correct the builder name soon :)

pablogsal: Just to confirm, what the correct name and tags should be by our convention?

zware: `tags=['linux', 'unix', 'arch', 'amd64', 'x86-64'],`

master/master.cfg
```diff
@@ -236,6 +236,7 @@ builders = [
     ("AMD64 Fedora Rawhide Clang Installed", "cstratak-fedora-rawhide-x86_64", ClangUnixInstalledBuild, UNSTABLE),
     ("AMD64 Fedora Rawhide LTO", "cstratak-fedora-rawhide-x86_64", LTONonDebugUnixBuild, UNSTABLE),
     ("AMD64 Fedora Rawhide LTO + PGO", "cstratak-fedora-rawhide-x86_64", LTOPGONonDebugBuild, UNSTABLE),
+    ("x86 Arch Linux", "pablogsal-arch-x86_64", UnixBuild, UNSTABLE),
```
zware: `("AMD64 Arch Linux", "pablogsal-arch-x86_64", UnixBuild, UNSTABLE),`

We pretty much use "AMD64" everywhere, although we have a macOS builder using "x86-64" (I suppose because it's definitely an Intel processor, but that also doesn't really matter anyway :)).

pablogsal: I have opened https://github.com/python/buildmaster-config/pull/127

PR #127, opened 2019-12-30
Correct arch linux worker name and tags

pablogsal:


pablogsal:
Thanks for the catch, @zware! I was a bit confused by the AMD64 part.

pablogsal:
I will deploy this change now :)

PR #129, opened 2020-01-04
Upload test result summaries to server for future analysis

pablogsal:


pablogsal:
Sadly, this PR will fail with the 2.7 branch as the junit option was not ported. When are we going to kill the 2.7 workers?

zware:
@benjaminp would prefer to keep the 2.7 workers until after the final release.  We can go ahead with this by guarding the extra options and step with a check of the branch.

Btw, there is `~buildbot/test-master` set up to avoid messing with the real master when trying out changes like this.  It's been a while since it ran, but it should basically work with a little bit of help :)

pablogsal:
@zware I have updated the PR by guarding the `2.7` branch.

master/custom/steps.py
```diff
@@ -106,3 +107,16 @@ class Uninstall(ShellCommand):
...
+class UploadTestResults(steps.FileUpload):
+    warnOnFailure = True
+    haltOnFailure = False
+    flunkOnFailure = False
+    alwaysRun = True
+
+    def __init__(self):
+        super().__init__(
+                workersrc="test-results.xml",
+                masterdest=util.Interpolate("~/test_results/%(prop:buildername)s_%(prop:buildnumber)s.xml")
```
zware: Should we make this a publicly-served directory?

pablogsal: > Should we make this a publicly-served directory?

What is the best directory that we should use for making it publicly-served?

zware: How about `/data/www/buildbot/test-results/`?  I just created it; it [seems to work](https://buildbot.python.org/test-results/).  Daily `.dmg` builds used to be uploaded to `/data/www/buildbot/daily-dmg/`.

pablogsal: Updated the PR!

master/custom/steps.py
```diff
@@ -106,3 +107,16 @@ class Uninstall(ShellCommand):
...
+class UploadTestResults(steps.FileUpload):
+    warnOnFailure = True
+    haltOnFailure = False
+    flunkOnFailure = False
+    alwaysRun = True
+
+    def __init__(self):
+        super().__init__(
+                workersrc="test-results.xml",
+                masterdest=util.Interpolate("/data/www/buildbot/test-results/%(prop:buildername)s_%(prop:buildnumber)s.xml")
```
zware: Perhaps a directory per builder?  Or even `<branch>/<builder>/build<number>.xml`?  This does have the potential to become quite crowded if left flat.

pablogsal: Done in 7e95c85

pablogsal:
Thanks for the review @zware ! 🎉

PR #130, opened 2020-01-04
Run buildbot with Python 3.7

pablogsal:
~ I have installed 3.8 in the server via pyenv~

We can not use 3.8 because it seems that the version of SSL is too old :/

Is there any possibility of upgrading the Ubuntu in the server?


pablogsal:
Closing this as the Ubuntu that we have in the server does not support anything newer than 3.6 :/

PR #131, opened 2020-01-05
Update macos builder name - this is being updated to Catalina

mattbillenstein:


pablogsal:
@mattbillenstein I will update the server and the local configuration.

Please, remember to update the call to `buildbot-worker create-worker'  to use your new worker name (`billenstein-macos`).

PR #132, opened 2020-01-06
Run buildbot with Python 3.8 - the future is NOW (now for real)

pablogsal:
I have compiled and installed 3.8 with LTO/PGO in the server using pyenv.


pablogsal:
Is alive! We live now in the future :)

PR #136, opened 2020-01-07
Upload results only if the test step fail

pablogsal:


vstinner:
Currently, the buildbot server only has 10 GB for data. We can ignore successful builds for now, until the partition is extended.

pablogsal:
> Currently, the buildbot server only has 10 GB for data. We can ignore successful builds for now, until the partition is extended.

👍

PR #140, opened 2020-01-07
Don't report missing workers via email

pablogsal:


zware:
I'm very against this; is this meant to be temporary?

pablogsal:
@zware Yes, is only while I figure out how to make test-master working. Sorry for the lack of clarity here.

PR #142, opened 2020-01-07
Allow to use local workers when there is no settings.yaml

pablogsal:
When working locally, we need to substitute the workers by `LocalWorker` so we can trigger builds as desired locally and also deactivate auth (so we don't need to log into the local instance using GitHub).

### DIFF
--- a/master/custom/workers.py
+++ b/master/custom/workers.py
@@ -11,7 +11,8 @@ class CPythonWorker:
 
     def __init__(self, settings, name,
                  tags=None, branches=None,
-                 parallel_builders=None, parallel_tests=None):
+                 parallel_builders=None, parallel_tests=None,
+                 local=False):
         self.name = name
         self.tags = tags or set()
         self.branches = branches
@@ -23,7 +24,10 @@ class CPythonWorker:
         pw = worker_settings.get('password', None) or owner_settings.password
         owner_email = owner_settings.get('email', None)
         emails = list(map(str, filter(None, (settings.get('status_email', None), owner_email))))
-        self.bb_worker = _worker.Worker(name, str(pw), notify_on_missing=emails)
+        if settings.value == ...:
+            self.bb_worker = _worker.LocalWorker(name)
+        else:
+            self.bb_worker = _worker.Worker(name, str(pw), notify_on_missing=emails)
 
 
 def get_workers(settings):
--- a/master/custom/workers.py
+++ b/master/custom/workers.py
@@ -24,7 +24,7 @@ class CPythonWorker:
         pw = worker_settings.get('password', None) or owner_settings.password
         owner_email = owner_settings.get('email', None)
         emails = list(map(str, filter(None, (settings.get('status_email', None), owner_email))))
-        if settings.value == ...:
+        if settings.use_local_worker:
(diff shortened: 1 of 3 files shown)


master/custom/workers.py
```diff
@@ -23,7 +24,10 @@ def __init__(self, settings, name,
         pw = worker_settings.get('password', None) or owner_settings.password
         owner_email = owner_settings.get('email', None)
         emails = list(map(str, filter(None, (settings.get('status_email', None), owner_email))))
-        self.bb_worker = _worker.Worker(name, str(pw), notify_on_missing=emails)
+        if settings.value == ...:
```
zware: I'd rather this be added as a setting in `settings`.

master/custom/workers.py
```diff
@@ -23,7 +24,10 @@ def __init__(self, settings, name,
...
+            self.bb_worker = _worker.LocalWorker(name)
+        else:
+            self.bb_worker = _worker.Worker(name, str(pw), notify_on_missing=emails)
```
zware: You can also just run whichever worker(s) you want separately rather than trying to run them all locally.

master/master.cfg
```diff
@@ -67,7 +67,7 @@ try:
     settings = Settings.from_file(settings_path)
 except FileNotFoundError:
     log.err(f'WARNING: settings file could not be found at {settings_path}')
-    settings = Settings()
+    settings = Settings({'do_auth': False})
```
zware: You can instead change the value in `settings.DEFAULTS`.

master/custom/workers.py
pablogsal: >You can also just run whichever worker(s) you want separately rather than trying to run them all locally.

Hummmm, you mean create the worker process separately? I find that a bit cumbersome, but if you think it makes more sense, we can do that.

pablogsal:
Will update the PR soon :)

Thanks for the review @zware !

master/custom/workers.py
zware: I'm just afraid that replacing `Worker` with `LocalWorker` means that it's going to try to run *all* of our defined workers locally, which is just going to lead to headache :).  But I'm perfectly fine with it being a (non-default) option.

pablogsal: > I'm just afraid that replacing `Worker` with `LocalWorker` means that it's going to try to run _all_ of our defined workers locally, which is just going to lead to headache :). But I'm perfectly fine with it being a (non-default) option.

Hum, maybe the correct thing to do is factor out the builders (because they have hardcoded worker names) into another module and create a `ExampleBuilder`. What do you think?

master/custom/workers.py
```diff
@@ -11,7 +11,8 @@ class CPythonWorker:
 
     def __init__(self, settings, name,
                  tags=None, branches=None,
-                 parallel_builders=None, parallel_tests=None):
+                 parallel_builders=None, parallel_tests=None,
+                 local=False):
```
zware: This new parameter doesn't seem to be used.

master/custom/workers.py
zware: I think what you've done with the `Test Builder` and `local-worker` are fine.

PR #143, opened 2020-01-08
Do not generate junit file on refleak builders

pablogsal:
This is temporary until we fix https://bugs.python.org/issue39253


pablogsal:
I am merging this to avoid an avalanche of failures with the refleak builders.

Issue #144, opened 2020-01-08
File permissions on uploaded test results are too restrictive


zware:
See [this result](https://buildbot.python.org/test-results/3.x/PPC64%20Fedora%203.x/build_145.xml), for example.

```
-rw------- 1 buildbot buildbot 4.6M Jan  8 15:54 '/data/www/buildbot/test-results/3.x/PPC64 Fedora 3.x/build_145.xml'
```

This wasn't true previously, see [this result](https://buildbot.python.org/test-results/3.x/PPC64%20Fedora%203.x/build_138.xml):

```
-rwxr-xr-x 1 buildbot buildbot 4.6M Jan  7 17:05 build_138.xml
```

This appears to be due to `umask` being set to 0077 in the current master process:

```
$ grep Umask /proc/`pgrep -f buildbot.tac`/status
Umask:  0077
```

pablogsal:
Seems that `FileUpload` has a `mode` option:

```
- ['mode']       file access mode for the resulting master-side file.
                 The default (=None) is to leave it up to the umask of
                 the buildmaster process.
```

What should be the desired permissions?

zware:
Something world-readable, or at least readable by `www-data` :).  Otherwise, I'm not too bothered.

PR #145, opened 2020-01-08
Use non-restrictive permissions when transferring test results

pablogsal:
Closes: #144


pablogsal:
755 is the current mode of the `test-results` folder:

```
buildbot@virt-et2yi0:/data/www/buildbot$ stat -c "%a %n" test-results/
755 test-results/
```

pablogsal:
I have updated the permissions of the existing files in `test-results` as well.

PR #146, opened 2020-01-08
Log exceptions to sentry

pablogsal:
Ernest has set up a PSF sentry project for us! 🎉

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -11,6 +11,7 @@
 # buildmaster. They are documented in docs/config.xhtml .
 
 import os
+import subprocess
 import sys
 
 from datetime import timedelta
@@ -22,6 +23,9 @@ from buildbot.plugins import reporters, util
 from buildbot import locks
 from buildbot.www.auth import NoAuth
 from twisted.python import log
+
+import sentry_sdk
+from sentry_sdk.integrations.sqlalchemy import SqlalchemyIntegration
 
 sys.path.append(os.path.dirname(__file__))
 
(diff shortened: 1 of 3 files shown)


master/master.cfg
```diff
@@ -65,12 +69,31 @@ from custom.workers import get_workers  # noqa: E402
 settings_path = os.path.join(os.path.dirname(__file__), 'settings.yaml')
 try:
     settings = Settings.from_file(settings_path)
+    try:
+        release_sha = subprocess.check_output(["git", "rev-parse", "HEAD"], text=True)
+        sentry_sdk.init(dsn=settings.sentry_dns, release=release_sha,
```
zware: `dsn` or `dns`?

This should probably use `settings.get('sentry_dns', None)` and check for `None`.

master/master.cfg
```diff
@@ -65,12 +69,31 @@ from custom.workers import get_workers  # noqa: E402
...
+                        integrations=[SqlalchemyIntegration()])
+    except subprocess.SubprocessError:
+        log.err('WARNING: Could not get current SHA for the release!')
```
zware: Is this meant to wrap the `init` as well?

master/master.cfg
```diff
@@ -65,12 +69,31 @@ from custom.workers import get_workers  # noqa: E402
...
+
+def logToSentry(event):
+    if not event.get('isError') or 'failure' not in event:
+        return
+
+    f = event['failure']
+    sentry_sdk.capture_exception((f.type, f.value, f.getTracebackObject()))
+
+
+log.addObserver(logToSentry)
```
zware: This should probably only be added if sentry was actually set up successfully.

master/master.cfg
pablogsal: Is DNS (Data Source Name). You can check in https://docs.sentry.io/error-reporting/quickstart/?platform=python

master/master.cfg
pablogsal: Nop, thanks for the catch!

master/master.cfg
```diff
@@ -62,13 +66,35 @@ from custom.settings import Settings    # noqa: E402
...
 
+
+def set_up_sentry():
+    try:
+        release_sha = subprocess.check_output(["git", "rev-parse", "HEAD"], text=True)
+    except subprocess.SubprocessError:
+        log.err('WARNING: Could not get current SHA for the release!')
+        return
+
+    sentry_sdk.init(dsn=settings.sentry_dns, release=release_sha,
```
zware: Same comments :).  `dsn` or `dns`; you have both in different spots here.  And this should check whether `settings.get('sentry_d..', None)` is `None` and act accordingly, issuing a warning if it *is* `None` and doing the setup otherwise.

pablogsal: Oh, sorry I totally missed that :( Thanks for the catch and the patience on the review!

pablogsal:
@zware Updated in 1e6fb2f

PR #148, opened 2020-01-14
[ON HOLD] Add a RHEL8 s390x worker

stratakis:


pablogsal:
@stratakis Do you have a worker password already?

stratakis:
I've got worker passwords for the rest of my buildbots (different for each), but notpassword has been generated for me since the new config was deployed. Can I reuse a pass from one of my workers or does a new one have to be generated?

zware:
With some recent changes to the config (namely the one that made the worker setup more public), you can have a single password for all workers or a unique password per worker.  You currently have both; your RHEL{7,8}-x86_64 and fedora-stable workers each have their own password while your fedora-rawhide worker is using your "owner" password.  We can generate a new one for this worker, or you can use the same one as fedora-rawhide as soon as this is deployed, whichever you would prefer.

stratakis:
Putting it on hold for the time being.

vstinner:
@stratakis: Contact me directly. I can provide you a password and merge this PR, once the worker is ready on your side.

stratakis:
Closing this in favor of #160

PR #152, opened 2020-01-19
Add release dashboard to the buildbot menu

pablogsal:
dsfsdf
Screenshot from 2020-01-19 19-36-11


pablogsal:
Screenshot from 2019-10-02 01-55-36

pablogsal:
Screenshot from 2020-01-19 19-36-11

zware:
I've not looked in detail because I'm not familiar with how most of this works :).  However, I'm in favor of the idea, so I'd say go for it when you're confident in the code.

pablogsal:
Thanks, @zware!

I tested it on the master test and we can always iterate in the future :)

PR #155, opened 2020-01-20
Add s390x RHEL8 worker and rename RHEL worker to RHEL7

edelsohn:


stratakis:
Would you be able to add some more build configs?

My current configs for RHEL7 and RHEL8: https://github.com/python/buildmaster-config/blob/master/master/custom/builders.py#L64

For the reference leak build the job needs to be added here as well: https://github.com/python/buildmaster-config/blob/master/master/custom/builders.py#L115


Also depending on the cores of each machine you can adjust the -j option for make etc such as: https://github.com/python/buildmaster-config/blob/master/master/custom/workers.py#L92

edelsohn:
I can add more builds. I would like to ensure that the basic build works and is stable first.

You can see the build times for the RHEL7 system on a normal build. They all should have 4 cpus and 16GB memory. I welcome advice about parallelism.

stratakis:
Then I guess a number of 6 for parallel tests should be good, by following the convention cpu cores + 2

master/master.cfg
```diff
@@ -190,7 +190,8 @@ builders = [
     ("PPC64LE Fedora", "edelsohn-fedora-ppc64le", UnixBuild, STABLE),
     ("s390x SLES", "edelsohn-sles-z", UnixBuild, STABLE),
     ("s390x Debian", "edelsohn-debian-z", UnixBuild, STABLE),
-    ("s390x RHEL", "edelsohn-rhel-z", UnixBuild, STABLE),
+    ("s390x RHEL7", "edelsohn-rhel-z", UnixBuild, STABLE),
```
stratakis: The change here is for the builder name but not the worker. Maybe also change the worker name? Not sure if it's gonna require any changes to the existing config in the worker though.

edelsohn: I specifically did not want to change the worker name and lose history.  I simply wanted to change the description to clarify versus the new RHEL8 worker.

stratakis: That is reasonable yeah

edelsohn:
Do you want me to generate a new pull request for the new refactorization?

edelsohn:
Cancelling pull request to rebase on refactored upstream repo.

PR #160, opened 2020-01-24
Add s390x RHEL8 worker

edelsohn:


pablogsal:
@edelsohn Do you want to go with the config as it is or you want to tweak the parallelism as per https://github.com/python/buildmaster-config/pull/155#issuecomment-576323375?

edelsohn:
Because the series of patches adds a large number of additional workers, I would like to do this methodically, incrementally and only change one thing at a time.

First, I would like to get a default worker for each Red Hat operating system variant, ensure that all prerequisite packages are installed, and achieve Green status. Then we can start to adjust the parallelism and add more builders with alternate configurations.

pablogsal:
@edelsohn 

Ok, I will land this and deploy, could you confirm (in 5 mins) that your new workers have a connection to the server?

PR #163, opened 2020-02-04
Add parallelism for the s390x workers

stratakis:


stratakis:
The workers have four cores, so this PR essentially adds the -j6 flag when running the tests.

stratakis:
@edelsohn

edelsohn:
This also can be done for the Debian Z and SLES Z buildbots, if you want.

stratakis:
I'll rebase. That should cut the build time significantly.

PR #164, opened 2020-02-07
Add ppc64le workers and builders

stratakis:


vstinner:
I deployed the new config manually.

PR #166, opened 2020-02-12
Add additional builders for the s390x workers

stratakis:


stratakis:
@edelsohn

stratakis:
Note that for the clang builds, the clang and compiler-rt packages need to be installed at the fedora workers.

edelsohn:
Clang and compiler-rt has been installed on RHEL8, Fedora 31, and Rawhide systems.
Clang-3.9 has been installed on the Debian system.

stratakis:
Note here that I haven't added the clang builds on the RHEL systems, as at least personally I haven't tested compiling python with clang on a rhel based distro, not sure how common that is.

Also I didn't add the clang builds on the Debian systems, but I could definitely add some more configs there if you would like. Note that the compiler-rt package is required for the [address sanitizer](https://clang.llvm.org/docs/AddressSanitizer.html) builds of clang which I have enabled for rawhide and fedora stable.

edelsohn:
compiler-rt is installed on RHEL8, Fedora, Rawhide.

stratakis:
So @edelsohn are you ok with adding those configs on the workers?

Summarizing here the additional builds that will be triggered on every commit:

On Fedora rawhide and Fedora stable:

Reference leaks
Clang with address sanitizer
Clang with installed python
Link time optimizations
Link time optimizations and Profile guided optimizations

On RHEL7 and RHEL8:

Reference leaks
Link time optimizations
Link time optimizations and Profile guided optimizations

edelsohn:
Let's give it a try and see what happens.  We'll monitor it and scale back the builds or scale up the resources if necessary.

vstinner:
I deployed the updated configuration. New s390x workers should pop up at https://buildbot.python.org/all/#/builders

PR #170, opened 2020-03-02
Have the refleak builds for aarch64 done once a day

stratakis:
On #169 I added the aarch64 workers and builders, including reference leak builds. However I forgot to add the refleak builds to the daily builder list, so those intensive builds are being run on every commit, hence I need to add them to the daily builders.


pablogsal:
Change deployed!

Issue #168, opened 2020-02-18
Disable Windows 7 on the Python 3.x branch


vstinner:
Can someone try to propose a PR on https://github.com/python/buildmaster-config/blob/master/master/custom/builders.py to remove Windows 7 from stable buildbots? That would remove them from stable on all Python branches.

The other option would be to hack https://github.com/python/buildmaster-config/blob/master/master/custom/workers.py to exclude 3.x branch from Windows 7 workers.

The last option is to hack https://github.com/python/buildmaster-config/blob/master/master/master.cfg to add a special rule to not schedule jobs on the 3.x branch if "win7" flag is present.

I'm not sure which approach is the best. Maybe @zware or @pablogsal have an opinion on that :-)

aeros:
In `builders.py`, it seems that the Windows XP builders are still present as well. Could we also remove those since XP is no longer supported by Microsoft?

See: https://github.com/python/buildmaster-config/blob/d87219bdd744ec59a52085e37d19dd6279e6aee0/master/custom/builders.py#L104-L112

zware:
We'll remove the XP worker and builders when we remove the 2.7 builders.

aeros:
Oh okay, I'll adjust the PR then.

aeros:
It was recently brought to my attention that this issue was already addressed by @vstinner in https://github.com/python/buildmaster-config/pull/171, so I'll close the issue. If there's something that still needs to be addressed, feel free to reopen it.

vstinner:
Windows XP is only used to test Python 2.7. See master.cfg:

```
        if "Windows XP" in name and branchname != "2.7":
            # 3.5+ drop support for XP
            continue
```

aeros:
@vstinner Was the intention to also disable Windows 7 when using the recently added `test-with-buildbots` label in PRs to 3.x? If so, I just saw a couple of Windows 7 builders present when I used it in https://github.com/python/cpython/pull/19149.

vstinner:
> @vstinner Was the intention to also disable Windows 7 when using the recently added test-with-buildbots label in PRs to 3.x? If so, I just saw a couple of Windows 7 buildbots present when I used it in python/cpython#19149 (AMD64 Windows7 SP1 PR and x86 Windows7 PR).

I proposed https://github.com/python/buildmaster-config/pull/176 to fix that.

PR #173, opened 2020-03-14
Remove unsupported Windows 7 from stable builders

aeros:
See https://github.com/python/buildmaster-config/issues/168.


zware:
@vstinner already removed the 3.x builders from the Windows 7 workers in #171; we don't want to change the "stable vs unstable" attribute of the Windows 7 workers.  They're still stable for older branches, and will stick around until those branches are out of support (or their owners remove them, but we're hoping for the former :)).

aeros:
@zware Ah, thanks for clarifying. I hadn't realized that Victor had already addressed it in another PR since there wasn't any mention of it in the issue: https://github.com/python/buildmaster-config/issues/168. Can the issue be closed?

PR #178, opened 2020-03-30
increase parallelization of PPC64 buildbot

### LINKED ISSUE #177: Increase parallelization of PPC64 builds
The server used for AIX PPC64 builds has 16 core (64 hardware threads) and can easily support more parallelization  during the bot tests.

aixtools:
Closes #177 
Increase  parallelization of PPC64 builds

### DIFF
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -226,7 +226,6 @@ class AIXBuild(UnixBuild):
         "--with-pydebug",
         "--with-openssl=/opt/aixtools",
     ]
-    testFlags = "-j8"
 
 
 class NonDebugUnixBuild(UnixBuild):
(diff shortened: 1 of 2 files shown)


aixtools:
Yes, looks like this may be a better way. I'll change the PR.

aixtools:
on edelsohn - as that will always be on POWER8, or better - and currently has 64 (logical) cpus. "aixtools" when on POWER6 only has 4 logical cpus - so leaving asis (for now).

aixtools:
We are a week further, and "sadly", imho - the keyword `parallel_tests` is not just test. As @isidentical mentioned - this also affects the parallelization of the build (`make -jXX`) and not just changing the argument for the test phase (make buildbottest).

If there is not a way to change only the argument to the `test phase` - reverting this change may be the better choice.

vstinner:
PRs are welcome, but I don't have the bandwidth to investigate this issue.

PR #182, opened 2020-04-13
Enable parallel compile & tests on gps-raspbian.

gpshead:


gpshead:
Looks like buildbot doesn't autodetect cpu & ram, manually specify the parallelization here. @zware

master/custom/workers.py
```diff
@@ -192,6 +192,8 @@ def get_workers(settings):
             name="gps-raspbian",
             tags=['linux', 'unix', 'raspbian', 'debian', 'armv6', 'armv7l',
                   'aarch32', 'arm'],
+            parallel_builders=5,
+            parallel_tests=7,
```
zware, line 196: This could allow up to 35 tests running at a time if we have 5 active branches all building; is that the intention?

gpshead: hah, nope.  I just want it `make all -j 5` and `python -m test.regrtest -j 7`.

i'm used to seeing buildbot serialize access to the bot.  if 3.x is building, 3.8 will be queued waiting until that one finishes, etc.  But appropriate build parallelism gets used within the individual builds.

The logs show serial make and -j 2 being passed to the tests right now.  that is a waste.

It is a four core machine, it needs 4+1 tasks.

gpshead: It looks like `custom/factories.py` needs a _new class defined_ *!!* and I have to update the `custom/builders.py` file to reference that instead?

sigh.  buildbot makes this so rediculious.

gpshead: master.cfg makes it look like setting _only_ `parallel_tests=6` will do what I want?  that one appears to get turned into a -j flag via a few levels of indirection.  Which then winds up passed as parallel="-j6" to the UnixBuilder factory specified in builders.py presumably winding up passed to setup where it finds it way to the make all and make buildbot test command line.

zware: Yes, things are a bit convoluted :(.  You did get it straightened out, though; sorry for not just laying that out in my previous comment!

PR #183, opened 2020-04-17
Add the aarch64 builders to the stable list

stratakis:
Move the aarch64 builders to the stable list for
the Fedora stable, RHEL7 and RHEL8 workers.

### DIFF
--- a/master/custom/builders.py
+++ b/master/custom/builders.py
@@ -81,20 +81,40 @@ def get_builders(settings):
         ("AMD64 RHEL8 Refleaks", "cstratak-RHEL8-x86_64", UnixRefleakBuild, STABLE),
         ("AMD64 RHEL8 LTO", "cstratak-RHEL8-x86_64", LTONonDebugUnixBuild, STABLE),
         ("AMD64 RHEL8 LTO + PGO", "cstratak-RHEL8-x86_64", LTOPGONonDebugBuild, STABLE),
+        # Linux PPC64le
         ("PPC64LE Fedora Stable", "cstratak-fedora-stable-ppc64le", UnixBuild, STABLE),
         ("PPC64LE Fedora Stable Refleaks", "cstratak-fedora-stable-ppc64le", UnixRefleakBuild, STABLE),
         ("PPC64LE Fedora Stable Clang", "cstratak-fedora-stable-ppc64le", ClangUbsanLinuxBuild, STABLE),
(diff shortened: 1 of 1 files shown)


stratakis:
The aarch64 buildbots have been for a long time now at the unstable list, however the builds are quite fine. Moving them to the stable list.

zware:
I think I'm ok with this; @vstinner or @pablogsal?  Should we ask the RMs' opinion?

pablogsal:
I am ok with this :ok_hand: 

> Should we ask the RMs' opinion?

IIRC other builders have been marked as stable before without RM notifications so not sure if is needed but on the other hand it wouldn't hurt.

vstinner:
> Move the aarch64 builders to the stable list for the Fedora stable, RHEL7 and RHEL8 workers.

Oh wow, that's a bunch of workers (14)!

The Refleak worker has a flacky test: https://bugs.python.org/issue39932

Can we start with a smaller set? I propose to start with 3 workers:
* "aarch64 Fedora Stable"
* "aarch64 RHEL7"
* "aarch64 RHEL8"

Watch then, and in 2 weeks, add a bunch of more workers.

To be honest, I get so many buildbot failures these days that I don't read them anymore. The latest email comes from "aarch64 Fedora Rawhide 3.x" :-p The second email comes from "aarch64 RHEL8 Refleaks 3.7" (another failure). So I wouldn't qualify "aarch64 RHEL8 Refleaks 3.7" as stable yet.

stratakis:
Makes sense to me. I'll close this and create a new PR.

> > Move the aarch64 builders to the stable list for the Fedora stable, RHEL7 and RHEL8 workers.
> 
> Oh wow, that's a bunch of workers (14)!
> 
> The Refleak worker has a flacky test: https://bugs.python.org/issue39932
> 
> Can we start with a smaller set? I propose to start with 3 workers:
> 
>     * "aarch64 Fedora Stable"
> 
>     * "aarch64 RHEL7"
> 
>     * "aarch64 RHEL8"
> 
> 
> Watch then, and in 2 weeks, add a bunch of more workers.

Makes sense to me. I'll close this and create a new PR.

stratakis:
> @stratakis Also be aware that once the builders are in the stable list, they become available for testing in GitHub when using the buildbot label

That is no problem, the buildbots are quite powerful so they should be able to handle the extra load, I should however add the specs information.

PR #187, opened 2020-04-22
Add a builder for the old parser

pablogsal:


master/custom/factories.py
```diff
@@ -115,6 +116,7 @@ def setup(self, parallel, branch, test_with_PTY=False, **kwargs):
                 command=["make", "pythoninfo"],
                 warnOnFailure=True,
             )
+            env=self.test_environ
```
vstinner, line 119: This syntax looks wrong. Do you want to pass it as an argument to ShellCommand?

It seems like env={} is the default: https://github.com/buildbot/buildbot/blob/master/master/buildbot/process/buildstep.py#L1129

pablogsal: > This syntax looks wrong. Do you want to pass it as an argument to ShellCommand?

The idea is that by default uses the default and if a class overrides it it uses de override

pablogsal: Check http://docs.buildbot.net/latest/manual/configuration/buildsteps.html

master/custom/builders.py
```diff
@@ -116,6 +117,7 @@ def get_builders(settings):
         ("AMD64 Fedora Rawhide LTO", "cstratak-fedora-rawhide-x86_64", LTONonDebugUnixBuild, UNSTABLE),
         ("AMD64 Fedora Rawhide LTO + PGO", "cstratak-fedora-rawhide-x86_64", LTOPGONonDebugBuild, UNSTABLE),
         ("AMD64 Arch Linux TraceRefs", "pablogsal-arch-x86_64", UnixTraceRefsBuild, STABLE),
+        ("AMD64 Arch Linux VintageParser", "pablogsal-arch-x86_64", UnixVintageParserBuild, STABLE),
```
vstinner, line 120: Hum, it may be safer to start with UNSTABLE, and only switch it to STABLE once we checked that it's stable.

pablogsal: > Hum, it may be safer to start with UNSTABLE, and only switch it to STABLE once we checked that it's stable.

Is the old parser, so it should be stable, no?

vstinner:
Hum, the worker got jobs for 3.7, 3.8 and master branches: you should maybe skip 3.7 and 3.8, no?

vstinner:
Good: the worker works as expected. test.pythoninfo says:

```
config[_use_peg_parser]: 0
```

pablogsal:
> Hum, the worker got jobs for 3.7, 3.8 and master branches: you should maybe skip 3.7 and 3.8, no?

Yeah, will do a PR soon for that.

pablogsal:
> Good: the worker works as expected. test.pythoninfo says:
> 
> ```
> config[_use_peg_parser]: 0
> ```

Thanks for checking

pablogsal:
> Hum, the worker got jobs for 3.7, 3.8 and master branches: you should maybe skip 3.7 and 3.8, no?

DOne in https://github.com/python/buildmaster-config/pull/188

PR #189, opened 2020-04-24
Add a FIPS enabled RHEL8 buildbot

stratakis:
Addition of RHEL 8 buildbot with [FIPS mode](https://en.wikipedia.org/wiki/FIPS_140-2) enabled.

This config is experimental, and Python will possibly not compile at all initially with it. Adding it to the unstable list, and feel free to turn off the notifications from it.


stratakis:
cc @tiran

vstinner:
New config deployed. Let's see how Python likes FIPS.

PR #191, opened 2020-04-29
Fix SlowUnixRefleakBuild timeout

vstinner:
Use a timeout of 5 hours rather than 30 minutes.


vstinner:
@pablogsal: oops, I missed that in your PR :-)

Maybe tests failed with timeout:
https://buildbot.python.org/all/#/builders/16/builds/161

pablogsal:
@vstinner Are you going to deploy?

vstinner:
> @vstinner Are you going to deploy?

Yeah, it's already deployed. Let's see tomorrow if I didn't mess up :)

PR #192, opened 2020-04-30
Add the normal clang build and replace some builders with it.

stratakis:


vstinner:
Since there are only few developers working on actually fixing UBSan issues, I'm fine with having less UBSan buildbot workers, as soon as we still have at least one UBSan worker. It's the case with this PR:

```
        ("AMD64 Clang UBSan", "gps-clang-ubsan", ClangUbsanLinuxBuild, UNSTABLE),
```

UBSan builds produce tons of warnings which make it harder to identify regular clang compiler warnings.

vstinner:
PR merged, new config deployed. Thanks @stratakis.

PR #193, opened 2020-05-11
Do not report back to GitHub with the PR buildeers

pablogsal:
It seems that the PR builders can get confused if a pull requests
contains commits of another pull request when ussing the
`test-with-buildbots` label. For instance, this PR:

https://github.com/python/cpython/pull/20006

included messages that report to a commit that lives in another PR where
the label was added. To avoid this, exclude the PR builders from the
GitHub notification system.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -312,8 +312,6 @@ for name, worker, buildfactory, stability in BUILDERS:
         )
     )
 
-github_status_builders.extend(pull_request_builders)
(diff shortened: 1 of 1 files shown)


vstinner:
What is the status of this PR?

pablogsal:
It should be merged

vstinner:
Thanks. Better late than never.

PR #198, opened 2020-05-19
Remove my x86 Gentoo Refleaks bulider

zware:
This box is starting to really show that it's becoming a teenager.


vstinner:
Thanks for this very useful Refleak worker! It saved me an infinite number of times!

PR #204, opened 2020-06-16
Update to buildbot v2.8.2

zware:


zware:
Travis-CI seems to have given up reporting back to us, so I disabled the requirement that it pass.  It *does* still run and pass, it just doesn't report back 🤷‍♂️

vstinner:
Thanks.

PR #221, opened 2020-11-16
bpo-42336: Updated PCbuild batch files

zooba:
The default settings in the PCbuild/*.bat files are being updated, so we'll need this update as well to maintain buildbots.

### DIFF
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -348,10 +348,10 @@ class WindowsBuild(TaggedBuildFactory):
     test_command = [r"Tools\buildbot\test.bat"]
     clean_command = [r"Tools\buildbot\clean.bat"]
     python_command = [r"python.bat"]
-    buildFlags = []
+    buildFlags = ["-p", "Win32"]
     remoteDeployFlags = []
     remotePythonInfoFlags = []
-    testFlags = ["-j2"]
+    testFlags = ["-p", "Win32", "-j2"]
     cleanFlags = []
     test_timeout = None
     factory_tags = ["win32"]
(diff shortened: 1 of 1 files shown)


zooba:
I'm not entirely sure how best to schedule the merges... I _guess_ we should merge this one first and let builds fail until the other one is merged? Or maybe it makes more sense to go the other way around. (They'll fail because I've also switched these to the more consistent `-p` flag, but I guess that could be split out into a later PR to try and avoid any breakage at all? But if we merge them all at the same time, no big deal, right?)

pablogsal:
> But if we merge them all at the same time, no big deal, right?)

Yeah, the worst case, if we do two PRs, is that 1 build will fail but all others will succeed and you will get some noise in the breaking PR from the reporter.  If you are able to merge botch changes into the same PR then everything should be green. I will be keeping an eye on the build bots in case we need to do something else but as long as we know about it, I would recommend to preceded in the way is more comfortable for you :)

pablogsal:
I think you can land in this repo, but otherwise, ping me and I will land it

zooba:
I can't merge here, so go ahead whenever you like and I'll merge https://github.com/python/cpython/pull/23275 and keep an eye on the buildbots. Looks like they're not busy now, so hopefully it'll be smooth enough.

vstinner:
> The default settings in the PCbuild/*.bat files are being updated, so we'll need this update as well to maintain buildbots.

Do you plan to update 3.8 and 3.9 as well? 3.9 yes according to the labels on your PR.

vstinner:
@pablogsal @zware:  it seems like we still support 3.7 in buildbots, no?

zooba:
I wasn't going to touch 3.7. Guess I can do 3.8, but I'll customise the patch to not change the default build (since the default download for 3.8 was 32-bit).

vstinner:
I'm asking if this PR should run a differerent command depending on the branch.

zooba:
It could... how do I make it do that?

zware:
> @pablogsal @zware: it seems like we still support 3.7 in buildbots, no?

I think we could remove 3.7 now, if @ned-deily is ok with it.

> I'm asking if this PR should run a differerent command depending on the branch.

I'd rather avoid that if we can, but it is an option.

pablogsal:
> I'd rather avoid that if we can, but it is an option.

I don't see other way if we are going to use options that are not available on 3.8 and 3.7

zooba:
https://github.com/python/cpython/pull/23325 will add the option on 3.8 without changing the defaults. It can probably backport safely to 3.7, too

ned-deily:
> It can probably backport safely to 3.7, too

If you feel up to doing that, I would be ok with merging to 3.7 (with the default as 32-bit). If the 3.7 buildbots are not taking needed resources needed by other branches, it would be good to keep buildbots for at least the most important platforms available to test security fixes for the branch.

zooba:
"Feel up to" means adding the tag and then pinging you to hit merge (I hope) ;)

I'm ready again to make this happen if someone merges the config.

pablogsal:
> I'm ready again to make this happen if someone merges the config.

Granted ✨

PR #228, opened 2021-01-26
Skip PGO builds on Python 3.7

vstinner:


stratakis:
The change seems sane.

PR #233, opened 2021-04-07
Promote the "AMD64 RHEL8 FIPS Only Blake2 Builtin Hash" builder to stable

stratakis:


stratakis:
The builder was failing with the traceback reported on https://bugs.python.org/issue43164

It seems that was an issue with the server that specific tests were connecting to and it has been fixed: https://buildbot.python.org/all/#/builders?tags=%2Bno-builtin-hashes-except-blake2

vstinner:
"AMD64 RHEL8 FIPS Only Blake2 Builtin Hash 3.9" is only green since 6 builds, and the latest one is orange (test_asyncio random failure, that's ok). I suggest to wait a little bit longer. For example, wait one week, to see if it's reliable.

At least, AMD64 RHEL8 FIPS Only Blake2 Builtin Hash 3.x looks very reliable: green from build 1022 to build 1054. The latest error was the build 1021 (test_nntplib).

By the way, do you know why test_nntplib failed and now pass?

vstinner:
> The builder was failing with the traceback reported on https://bugs.python.org/issue43164

Ah, that was the test_nntplib failure. Do you know why it failed and now it pass?

stratakis:
> > The builder was failing with the traceback reported on https://bugs.python.org/issue43164
> 
> Ah, that was the test_nntplib failure. Do you know why it failed and now it pass?

I can only assume that it was an issue with the server configuration which they fixed on their end. The domain is nntp.aioe.org and their website is http://news.aioe.org/ however I couldn't find any sort of changelog or news related to that matter.

stratakis:
Also this PR can wait as the buildbot is green for some days only, so we can wait a bit to verify that it's indeed stable.

stratakis:
Unfortunately two tests fail now on this buildbot after the recent addition of some tests. Fix on the main cpython repo for it: https://github.com/python/cpython/pull/26470

pablogsal:
@stratakis ping me once the buildbot is green so we can merge

stratakis:
The buildbot is green again.

PR #234, opened 2021-04-10
Added new build workers for Windows

jkloth:


jkloth:
Adding recent contributors and buildbot watchers @pablogsal @vstinner

Note that I have already switched the hardware for my main worker, but have left the newly added (renamed) worker offline so as to not flood the build master with connection requests.

pablogsal (approved):

pablogsal:
@jkloth Do you have already keys(passwords) for those builders I assume, no?

jkloth:
I believe that the password stays the same for all workers from the same owner, given the format '{owner}-{worker}' for worker naming.

pablogsal:
> I believe that the password stays the same for all workers from the same owner, given the format '{owner}-{worker}' for worker naming.

Yup, but some owners have several ones for different workers so I just wanted to check :)

Fidget-Spinner:
I'm not sure if this is the correct place to comment (sorry if it isn't, I'm not too familiar with the buildbots in CPython), but recently any keyword changes in the new worker have failed to compile. For example: https://buildbot.python.org/all/#/builders/596/builds/2/steps/2/logs/stdio

It seems that the regen grammar script needs >Python 3.8 as it's using the walrus, while the worker detects and runs ``py -3.7`` on the script.

This wasn't triggered earlier because that specific part of the script isn't run if no keyword changes are detected.

pablogsal:
> It seems that the regen grammar script needs >= Python 3.8 as it's using `:=`, while the worker detects and runs `py -3.7` on the script.

@jkloth Could you adapt your buildbot so Python3.8 is used for regenerating files as is done in the UNIX Makefile?

jkloth:
This worker has both 3.9 and 3.7 (both 64-bit) installed.  Running `py -0p` from the terminal shows that 3.9 is also the default.  That would mean that there is a bug in the build scripts for choosing the incompatible python installation.

pablogsal:
> This worker has both 3.9 and 3.7 (both 64-bit) installed. Running `py -0p` from the terminal shows that 3.9 is also the default. That would mean that there is a bug in the build scripts for choosing the incompatible python installation.

Hummmm, interesting. But we have Windows builders in the CI and those are ok with it. In fact, this seems to be the only buildbot that is failing with this.

Fidget-Spinner:
I created the draft PR https://github.com/python/cpython/pull/25389 to see if it fixes this.

PR #237, opened 2021-04-20
Use ws_ping_interval to avoid disconnects

pablogsal:


master/master.cfg
```diff
@@ -379,6 +379,7 @@ c["www"] = dict(
         },
     },
     plugins=dict(waterfall_view={}, console_view={}, grid_view={}),
+    ws_ping_interval=30,
```
vstinner, line 382: Maybe add a comment to https://bugs.python.org/issue41701

pablogsal: I still need to figure out how to update the server

pablogsal:
I'm going to merge this to try out if this fixes the problem, we can revert it later if it doesn't

pablogsal:
Unfortunately, there are several problems with buildbot 3.10.1+ so I had to downgrade: https://github.com/python/buildmaster-config/pull/240 :(

PR #240, opened 2021-04-21
Downgrade to buildbot 2.10.1

pablogsal:
Apparently, there are several bugs with the waterfall plugin, so we need to downgrade for the time being.


pablogsal:
I managed to fix the issue, so I have reverted this PR back. Sorry for the noise, but this has been quite hard to troubleshoot in the new server :(

PR #244, opened 2021-05-03
Update references to master to point to main

pablogsal:


vstinner:
See in master/master.cfg:

```
        # 2019-06-19: Test suite - AIX passes on master and 3.X branches, fails on 3.7
...
            # Workers known to be broken on older branches: let's focus on
            # supporting these platforms in the master branch. Don't run
            # custom jobs neither.
```

vstinner (approved):
🤖 Victor happy  🤖

PR #246, opened 2021-05-16
Aix ppc64 xlc builder

edelsohn:
Python on AIX is available in a version built with IBM XLC and a version built with GCC, so builders for each configuration would be beneficial.

This patch updates the Python Buildbot configuration to add a factory to use the IBM XLC compiler and a builder on the edelsohn-aix-ppc64 worker to use the new factory.


pablogsal (approved):

pablogsal:
Thanks @edelsohn !

PR #248, opened 2021-05-25
IRC: enable colors

vstinner:


vstinner:
https://docs.buildbot.net/latest/manual/configuration/reporters/irc.html#irc-bot

PR #249, opened 2021-05-26
Switch from Freenode to Libera chat

vstinner:


pablogsal:
> This PR is not enough, where is the secret settings file? How I can modify it?
> 
> cc @pablogsal @zware 

Is on the buildbot server in /etc/buiidbot/something IIRC.

vstinner:
I modified /etc/buildbot/settings.yaml and registered cpython-buildbot nickname there.

vstinner:
> Is on the buildbot server in /etc/buiidbot/something IIRC.

I documented it: https://github.com/python/buildmaster-config/blob/main/README.md#private-settings

PR #255, opened 2021-07-08
Use Git "fresh" method

vstinner:
Git "clean" method keeps most files created by a previous build. Use
Git "fresh" method instead to ignores .gitignore rules and so remove
all generated files (like ".o" files): run a fresh build rather than
an incremental build.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -227,8 +227,10 @@ for branch_num, (git_url, branchname, git_branch) in enumerate(git_branches):
             repourl=git_url,
             branch=git_branch,
             timeout=3600,
+            # "git clean -fdx": remove all files which are not tracked by Git,
+            # ignoring the .gitignore rules (ex: remove also ".o" files).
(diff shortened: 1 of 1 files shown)


vstinner:
cc @pablogsal @zware 

See https://docs.buildbot.net/latest/manual/configuration/steps/source_git.html documentation.

vstinner:
This change should fix https://mail.python.org/archives/list/buildbot-status@python.org/thread/Z27NUT4OY7IGUNQXCTMO4GGSR7RWVB4J/ issue.

Copy of my message:



setup.py failed with:

"RuntimeError: subprocess not supported for isolated subinterpreters"

It seems like aarch64 Fedora Rawhide 3.x doesn't use a fresh build,
but incremental build. I get such error when I use incremental build.

How can I check if the build is incremental or not? How can it be changed?

Victor

pablogsal (approved):

PR #256, opened 2021-07-19
Increase parallelism for the aarch64 RHEL and Fedora workers

stratakis:
Due to a hardware upgrade, each worker now boasts 40 cores
hence increasing the tests run in parallel here.


pablogsal (approved):

vstinner:
Yeah! Parallelize all the things!

PR #261, opened 2021-08-17
Allow python-triage to stop builds

### LINKED ISSUE #260: Should triagers be able to stop builds?
Hi, Python triagers can add `test-with-buildbots` label to PRs, but they cannot stop a build if it hangs. Currently, only core devs can do that. This makes it hard for triagers to "unstick" a PR after they test it with buildbots.

I think triagers should be able to stop builds. I will open a discussion at https://discuss.python.org/c/core-workflow/8. Please let me know if I should open a discussion at python-buildbots/python-dev too. Thanks for your time.

pablogsal:
closes #260

### DIFF
--- a/master/custom/auth.py
+++ b/master/custom/auth.py
@@ -22,13 +22,18 @@ def set_up_authorization(settings):
                 util.StopBuildEndpointMatcher(
                     role="buildbot-owners", defaultDeny=False
                 ),
+                util.StopBuildEndpointMatcher(role="python-triage", defaultDeny=False),
                 util.StopBuildEndpointMatcher(role="python-core"),
                 util.RebuildBuildEndpointMatcher(role="owner", defaultDeny=False),
+                util.RebuildBuildEndpointMatcher(
+                    role="python-triage", defaultDeny=False
+                ),
                 util.RebuildBuildEndpointMatcher(
                     role="buildbot-owners", defaultDeny=False
                 ),
                 util.RebuildBuildEndpointMatcher(role="python-core"),
                 util.ForceBuildEndpointMatcher(role="owner", defaultDeny=False),
+                util.ForceBuildEndpointMatcher(role="python-triage", defaultDeny=False),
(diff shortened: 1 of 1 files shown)


pablogsal:
closes #260

pablogsal:
CC @Fidget-Spinner

master/custom/auth.py
```diff
@@ -22,6 +22,7 @@ def set_up_authorization(settings):
                 util.StopBuildEndpointMatcher(
                     role="buildbot-owners", defaultDeny=False
                 ),
+                util.StopBuildEndpointMatcher(role="python-triage"),
```
zware, line 25: I *think* you need `defaultDeny=False` here.  I'm not certain of that, it's been quite a while since I read the docs for this.

While we're in here, we might as well allow triagers to rebuild or force, too.

pablogsal: Thanks for the review. I have addressed it :)

master/custom/auth.py
```diff
@@ -22,13 +22,15 @@ def set_up_authorization(settings):
                 util.StopBuildEndpointMatcher(
                     role="buildbot-owners", defaultDeny=False
                 ),
+                util.StopBuildEndpointMatcher(role="python-triage", defaultDeny=False),
                 util.StopBuildEndpointMatcher(role="python-core"),
                 util.RebuildBuildEndpointMatcher(role="owner", defaultDeny=False),
                 util.RebuildBuildEndpointMatcher(
                     role="buildbot-owners", defaultDeny=False
                 ),
```
zware, line 30: Add rebuild for triage?

zware (approved):

Fidget-Spinner:
Thanks Zach and Palo!

vstinner:
@Fidget-Spinner: This permission should maybe be listed in https://devguide.python.org/triaging/#becoming-a-member-of-the-python-triage-team

Fidget-Spinner:
@vstinner sure, I will work on that.

PR #262, opened 2021-09-03
Increase the timeout of slow buildbots

vstinner:
On x86 Gentoo Non-Debug with X, test_peg_generator was killed after
25 minutes, it took 22 minutes on a previous successful build.


vstinner:
@pablogsal @zware: FYI I changed the project to use "Squash and merge", I'm not a big fan of merge commits.

PR #263, opened 2021-09-03
Replace Travis CI with GitHub Action

vstinner:


vstinner:
@pablogsal @zware: It seems like .travis.yml was no longer used. So I switch to GitHub Action instead.

PR #264, opened 2021-09-09
Add FedoraBuild for Fedora and RHEL

vstinner:
Test ./configure --with-platlibdir=lib64 on Fedora and RHEL, but also
other configure flags used by the Fedora specfile, like
--enable-shared and --with-system-expat.


vstinner:
>  Build / build (pull_request) Failing after 14s — build 

"make" when installing Tempita in a venv:

```
error in Tempita setup command: use_2to3 is invalid.
```

See https://twitter.com/mitsuhiko/status/1435661088746577930

"Latest setuptools removes the ability to invoke 2to3 as part of the installation. With that Flask-OpenID broke which means that Airflow broke. I don't even have the energy to fix an ancient package at this point but apparently existing airflow installations are now screwed :-/"

flask-openid fix to avoid 2to3: https://github.com/mitsuhiko/flask-openid/pull/60

vstinner:
For Tempita use_2to3 error, see: https://bugs.launchpad.net/sqlalchemy-migrate/+bug/1943127

vstinner:
I also reported the issue to buildbot: https://github.com/buildbot/buildbot/issues/6229

vstinner:
`error in Tempita setup command: use_2to3 is invalid.`

I pushed a workaround: https://github.com/python/buildmaster-config/commit/73a70b5e0a94d4c94e518c5d1f9b02f691b23e3d

master/custom/factories.py
```diff
@@ -363,6 +363,34 @@ class NoBuiltinHashesUnixBuildExceptBlake2(UnixBuild):
...
+    configureFlags = [
+        "--with-pydebug",
+        "--with-platlibdir=lib64",
+        "--enable-ipv6",
+        "--enable-shared",
+        "--with-computed-gotos=yes",
+        "--with-dbmliborder=gdbm:ndbm:bdb",
+        "--with-system-expat",
+        "--with-system-ffi",
+        "--with-system-libmpdec",
```
hroncok, line 381: Adding options like this means we need to pull in the devel packages. Also, IIRC we can only use system mpdec in 3.10+

master/custom/factories.py
```diff
@@ -363,6 +363,34 @@ class NoBuiltinHashesUnixBuildExceptBlake2(UnixBuild):
...
 
+class FedoraBuild(UnixBuild):
+    # Build on Fedora or RHEL.
+    #
+    # Try to be as close as possible to the Fedora specfile used to build
+    # the RPM package:
+    # https://src.fedoraproject.org/rpms/python3.10/blob/rawhide/f/python3.10.spec
+    configureFlags = [
+        "--with-pydebug",
+        "--with-platlibdir=lib64",
```
hroncok, line 374: This only makes sense on 64bit arches.

master/custom/factories.py
stratakis: The devel packages are already in the buildbots

master/custom/factories.py
stratakis: All the Fedora and RHEL buildbots are 64 bits.

vstinner:
@hroncok @stratakis: I prefer to start easy, I disabled --with-system-expat, --with-system-ffi and --with-system-libmpdec. Let's see first how many buildbot workers are broken by this first incomplete change :-D

Also, I created one configuration per OS version:

* Fedora Stable
* Fedora Rawhide
* RHEL7
* RHEL8

For now, RHEL7 and RHEL8 are the same, so are Fedora Stable and Fedora Rawhide.

vstinner:
s390x RHEL8 3.10: configure failed: https://buildbot.python.org/all/#/builders/700/builds/341

```
checking for valgrind/valgrind.h... no
configure: error: Valgrind support requested but headers not available
```

PPC64LE RHEL7 3.10: https://buildbot.python.org/all/#/builders/619/builds/311

(same Valgrind error)

 s390x RHEL7 3.10:  https://buildbot.python.org/all/#/builders/658/builds/351

(same Valgrind error)

s390x Fedora 3.10: https://buildbot.python.org/all/#/builders/655/builds/338


(same Valgrind error)

hroncok:
That is because Valgrind has:

```
ExclusiveArch: %{ix86} x86_64 ppc ppc64 ppc64le s390x armv7hl aarch64
```

https://src.fedoraproject.org/rpms/valgrind/blob/rawhide/f/valgrind.spec#_230

Fedora's Python has:

```
# Support for the Valgrind debugger/profiler
%ifarch %{valgrind_arches}
%bcond_without valgrind
%else
%bcond_with valgrind
%endif
```

vstinner:
I rebuilt "s390x RHEL8 3.10" without the Valgrind flag and it went a little bit further until...

```
checking for dtrace... not found
configure: error: dtrace command not found on $PATH
```

vstinner:
I disabled Valgrind: https://github.com/python/buildmaster-config/commit/d05087c24d8c2bd94a64a642acc409f4d9771149

And DTrace: https://github.com/python/buildmaster-config/commit/a013e1783e30946249a0640ba7649ce87380fad9

PR #271, opened 2021-09-13
Update requirements

vstinner:


vstinner:
New versions deployed.

PR #289, opened 2021-10-03
Add Windows 11 build worker

jkloth:


pablogsal (approved):

pablogsal:
Wow, nice! Thanks for the PR @jkloth .

I am going to deploy the changes. Could you confirm that the builder is operative and running?

jkloth:
Up and running (and failing as expected :)

-- 
Jeremy Kloth

pablogsal:
> (and failing as expected :)

Ah, I though this was expected to succeed. In that case I will mark it as UNSTABLE for now

vstinner:
> Ah, I though this was expected to succeed. In that case I will mark it as UNSTABLE for now

In general, I would suggest to always start with UNSTABLE, and only mark as STABLE after 1 week of successful builds.

jkloth:
> Ah, I though this was expected to succeed. In that case I will mark it as UNSTABLE for now

It would succeed once https://bugs.python.org/issue45354 (submitted at the same time as the buildbot) is resolved.  It seems there is a shortage of developers willing/able to handle Windows-only items.

pablogsal:
> It seems there is a shortage of developers willing/able to handle Windows-only items.

Indeed. Maybe you can convince @zooba to take a look :)

pablogsal:
Also, one of the problems is to get a Windows 11 machine :(

pablogsal:
I would try to install it on my main computer and take a look but there is all the TMP shenanigans and I don't want to deal with that right now.

jkloth:
> Also, one of the problems is to get a Windows 11 machine :(

That was my point in getting this buildbot up and running prior to the public release of Win11 on Oct 5th

PR #290, opened 2021-10-11
Mark Windows ARM64 builds as Stable

niyas-sait:
[Win/ARM64 worker](https://buildbot.python.org/all/#/workers/45) has been running for around 10 days and seems to be stable


nsait-linaro:
@pablogsal Could you please review this change ?

pablogsal (approved):

PR #291, opened 2021-10-15
Add Oracle Solaris SPARC build worker

kulikjak:
I cloned this repo and tested everything internally. There are a few failing tests, but some of those have open PRs with fixes already and the rest can be fixed once the worker is running (right?).


vstinner:
If you are working on fixing remaining tests, I'm fine with adding the buildbot. If it becomes too complex to fix remaining tests, FYI I'm also fine with skipping them on Solaris.

PR #296, opened 2021-11-29
Mark AMD64 Debian root builder as unstable

pablogsal:


master/custom/builders.py
```diff
@@ -59,7 +59,7 @@ def get_builders(settings):
     return [
         # -- Stable builders --
         # Linux
-        ("AMD64 Debian root", "angelico-debian-amd64", UnixBuild, STABLE),
```
erlend-aasland, line 62: Also move it down to the unstable section?

pablogsal: This is supposed to be a temporary change, so I prefer to keep it here as hopefully we will put it back soon

pablogsal:
@tiran

PR #297, opened 2021-12-01
Revert "Mark AMD64 Debian root builder as unstable"

pablogsal:
Reverts python/buildmaster-config#296 now that the builder is stable again:

https://buildbot.python.org/all/#/builders/345


vstinner:
Nice, it's good to see it back online and green!

PR #300, opened 2021-12-15
Enable the out-of-tree builds for all the Fedora and RHEL buildbots

stratakis:


stratakis:
@vstinner

vstinner:
Let's do that. It's easy to revert if something goes wrong.

PR #302, opened 2022-01-26
Add CentOS Stream 9 workers and builders

stratakis:


master/custom/workers.py
```diff
@@ -135,9 +155,9 @@ def get_workers(settings):
             parallel_tests=40,
         ),
         cpw(
-            name="cstratak-RHEL8-fips-x86_64",
```
vstinner, line 138: Did you remove this config on purpose?

stratakis: Moved it up together with the other x86_64 ones.

vstinner:
Merged, thank you ;-)

vstinner:
test_json and test_lib2to3 started to crash on s390x Fedora Clang 3.x buildbot: https://bugs.python.org/issue46542

Can it be related to this change?

PR #311, opened 2022-02-01
Update requirements

vstinner:


vstinner:
Superset of https://github.com/python/buildmaster-config/pull/310

vstinner:
@pablogsal: You blocked the main branch of Python because of buildbots. Maybe today is a bad day to attempt upgrading requirements :-)

pablogsal:
Yeah, also, we won't be using this version in the speed.python.org server as we would need to remove all datapoints and the faster CPython team needs the old one (until we manage to get a version of the django server that knows how to deal with this).

vstinner:
> Yeah, also, we won't be using this version in the speed.python.org server as we would need to remove all datapoints and the faster CPython team needs the old one (until we manage to get a version of the django server that knows how to deal with this).

This PR is for buildbots, not benchmarks.

vstinner:
I recreated the venv on the server. I had to run ``venv/bin/buildbot upgrade-master /srv/buildbot/master`` to upgrade the buildbot database. And... the upgraded server is back online!

PR #316, opened 2022-03-18
Update Solaris worker configuration

kulikjak:
This slightly updates the configuration for the Solaris worker.

There are some weird issues in Python 3.8 killing its own master connection during tests, but newer runtimes seem not to do that, so let's just skip 3.8 (we are not using anyway) and focus on those.

Also, I have a question regarding the configuration: what is the `branches` argument in worker configuration for? It seems like it's not used anywhere and doesn't have any effect (which is why I had to skip 3.8 directly in the `master.cfg` file).

Lastly, I also opened https://bugs.python.org/issue47058 that should fix/skip tests due to which the test suite is failing.


vstinner:
Merged, thanks

PR #326, opened 2022-06-03
Add tier labels corresponding to PEP 11

brettcannon:


pablogsal (approved):

pablogsal:
Thanks @brettcannon! :heart:

pablogsal:
I merged so we have a baseline for tier support tags, but I noticed there are a bunch of Linux buildbots that are tagged as `NO_TIER` but they technically qualify as tier 1. Was this an oversight or do you want buildbot owners to manually tag those?

vstinner:
Great :-)

brettcannon:
I purposefully overlooked them as PEP 11 specifies the platforms run by CI as [tier 1](https://peps.python.org/pep-0011/#tier-1) and purposefully doesn't mention buildbots for that tier (hence why there isn't a single tier 1 buildbot 😁). That way the "Changes which would break the main branch are not allowed to be merged" requirement can be met since buildbots don't run on PRs.

If people think that's too strict of an interpretation and want to add tier 1 buildbots I'm fine with it.

PR #329, opened 2022-06-19
Don't use pydebug on WASM bots, reduce parallel tests

tiran:
Some tests are crashing or timing out. Test host is overloaded. Reduce
parallel tests and disable pydebug.

### DIFF
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -591,10 +591,6 @@ class WindowsARM64ReleaseBuild(WindowsARM64Build):
 
 
 class UnixCrossBuild(UnixBuild):
-    configureFlags = [
-        "--with-pydebug",
-        "--with-build-python=../build/python"
-    ]
     extra_configure_flags = []
     host_configure_cmd = ["../../configure"]
     host = None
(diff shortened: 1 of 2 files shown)


master/custom/factories.py
```diff
@@ -769,6 +766,8 @@ class Wasm32EmscriptenBuild(UnixCrossBuild):
 class Wasm32EmscriptenNodeBuild(Wasm32EmscriptenBuild):
     buildersuffix = ".emscripten-node"
     extra_configure_flags = [
+        # don't with --with-pydebug, Emscripten has limited stack
```
ezio-melotti, line 769:
```suggestion
        # don't run with --with-pydebug, Emscripten has limited stack
```

master/custom/factories.py
```diff
@@ -778,6 +777,8 @@ class Wasm32EmscriptenNodeBuild(Wasm32EmscriptenBuild):
 class Wasm32EmscriptenBrowserBuild(Wasm32EmscriptenBuild):
     buildersuffix = ".emscripten-browser"
     extra_configure_flags = [
+        # don't with --with-pydebug, Emscripten has limited stack
```
ezio-melotti, line 780:
```suggestion
        # don't run with --with-pydebug, Emscripten has limited stack
```

PR #332, opened 2022-07-01
Mark WASM buildbots as stable

tiran:
The wasm32-emscripten and wasm32-wasi build bots have been stable for a while. Mark them as stable.

image

### DIFF
--- a/master/custom/builders.py
+++ b/master/custom/builders.py
@@ -213,10 +213,10 @@ def get_builders(settings):
         ("ARM64 Windows Non-Debug Azure", "linaro2-win-arm64", WindowsARM64ReleaseBuild, UNSTABLE, NO_TIER),
 
         # WebAssembly
-        ("wasm32-emscripten node (pthreads)", "bcannon-wasm", Wasm32EmscriptenNodePThreadsBuild, UNSTABLE, NO_TIER),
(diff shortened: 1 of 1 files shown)


pablogsal (approved):

pablogsal:
Here we go! 🚀

vstinner:
@tiran: Recently, I got a few buildbot failures. I ignored them since you was tracking these WASM workers and there was work-in-progress. Since they are now marked as STABLE, I will pay more attention to them. If they are too many (random) failures, I suggest to move back affected workers to UNSTABLE until it's solved.

@tiran: I will keep in touch anyway, and I will not change the buildbot status (stable/unstable) without asking you ;-)

I'm mostly using https://mail.python.org/archives/list/buildbot-status@python.org/ to track worker failures.

vstinner:
@tiran @ethanhs @erlend-aasland and others: Congrats! Sorry my previous message is not very positive. In fact, I'm very happy of what's happening on Python 3.11/3.12 with WebAssembly! It's exciting!

tiran:
No problem! The WASM buildbots have been stable over the last weeks. We'll get even better hardware soon. On two, three occasions a new commit broke the builds. One time an update to EMSDK 3.1.14 caused failures, too. 3.1.15 is fine again. I'll keep an eye on the buildbots.

PR #335, opened 2022-07-25
Remove ware-win81-release worker

zware:
It's not coming back at this point, replaced by ware-win11.  If at some
point I get an 8.1 worker running again, I will re-add.


vstinner:
Bye and thanks little Windows machine!

PR #337, opened 2022-07-27
Use wasi-env script to set build parameters

tiran:
Instead of hard-coding build flags in the build bot configuration,
wasm32-wasi buildbot now uses CPython's wasi-env script. This allows us
to control the build parameters from CPython.

Depends on https://github.com/python/cpython/pull/95320 + backport to
3.11.


tiran:
CPython PR has landed in main and will land in 3.11 shortly.

PR #338, opened 2022-08-10
Ensure the email formatter has the logs content

pablogsal:
We were failing to construct the email because the log content
was missing and this was making it fail to fill the template.

### DIFF
--- a/master/custom/email_formatter.py
+++ b/master/custom/email_formatter.py
@@ -58,6 +58,7 @@ MESSAGE_FORMATTER = CustomMessageFormatter(
     template=MAIL_TEMPLATE,
     template_type="plain",
     want_logs=True,
+    want_logs_content=True,
     want_properties=True,
     want_steps=True,
 )


pablogsal:
> want_logs_content
This parameter (defaults to False) will extend the content of the logs with the log contents of each steps from the build. This implies want_logs and wantSteps to be True. Use it only when mandatory, as this greatly increases the overhead in terms of CPU and memory on the master.

vstinner:
> We were failing to construct the email because the log content was missing and this was making it fail to fill the template.

I don't get the rationale but I'm happy that emails will be back :-)

orsenthil:
Yeah, me too. Cannot connect this change to email being off / on. Especially turning it On.

pablogsal:
In the previous buidbot version before @vstinner updated it, the log contents were in the payload that we analyze to extract failures but in the new version, a backwards incompatible change forces us to add "want_log_content" so they are present in the payload again:

http://docs.buildbot.net/latest/manual/configuration/report_generators/formatter_function.html#messageformatterfunction

The problem is that when we updated the buildbot version because the content was not there we failed to format the message and they never got sent.

pablogsal:
The log here is referring to the BUILD logs (the output of the test suite).

PR #339, opened 2022-09-09
Promote Raspbian to stable, disable it for 3.8.

gpshead:
Disables it for the 3.8 branch as infrastructure changes early this year make `test_gdb` fail in that old branch; not important.

The bot lives on my home network which has battery backup but single source of failure connectivity and long term power outages would still be an issue. If it ever becomes a problem, we can demote it.


vstinner:
It seems like builds on Python 3.9, 3.10, 3.11 and main are successfull for 2 months, so I'm fine with marking the worker as stable.

PR #347, opened 2022-11-02
Run make regen-requirements

vstinner:


vstinner:
Fix https://github.com/python/buildmaster-config/security/dependabot/7

PR #348, opened 2022-11-08
Uninstall step: set warnOnFailure to True

vstinner:


vstinner:
Recent example: https://buildbot.python.org/all/#/builders/531/builds/2842

On "s390x Fedora Clang Installed 3.x", the ``git fetch -f -t https://github.com/python/cpython.git main --progress`` command failed with:

```
fatal: unable to access 'https://github.com/python/cpython.git/': The requested URL returned error: 503
```

The "git" step fails but has ``warnOnFailure = True``. Problem: the "uninstall" also failed because the target/ directory doesn't exist. It marked the whole step as a failure, whereas it's expected that uninstall fails. It's not a bug in Python, just a network issue.

PR #350, opened 2022-11-18
Add ambv-bb-win11 bigmem buildbot

ambv:


zware (approved):
LGTM

master/custom/workers.py
```diff
@@ -281,4 +281,10 @@ def get_workers(settings):
             parallel_tests=2,
             parallel_builders=2,
         ),
+        cpw(
+            name="ambv-bb-win11",
+            tags=['windows', 'win11', 'amd64', 'x86-64', 'bigmem'],
+            branches=['3.9', '3.10','3.11', '3.x'],
+            parallel_tests=8,
```
ambv, line 288: How is this used?

master/custom/factories.py
```diff
@@ -557,6 +557,15 @@ class Windows64Build(BaseWindowsBuild):
     factory_tags = ["win64"]
 
 
+class Windows64BigmemBuild(BaseWindowsBuild):
+    buildersuffix = ".bigmem"
```
ambv, line 561: What's that for?

master/custom/factories.py
```diff
@@ -557,6 +557,15 @@ class Windows64Build(BaseWindowsBuild):
...
+    buildFlags = ["-p", "x64"]
+    testFlags = ["-p", "x64", "-j8", "-M24g", "-u-all"]
```
ambv, line 563: I added `-j8` here, does that duplicate `parallel_tests` or should one of them be a different value?

master/custom/workers.py
```diff
@@ -281,4 +281,10 @@ def get_workers(settings):
             parallel_tests=2,
             parallel_builders=2,
         ),
+        cpw(
+            name="ambv-bb-win11",
+            tags=['windows', 'win11', 'amd64', 'x86-64', 'bigmem'],
+            branches=['3.9', '3.10','3.11', '3.x'],
```
ambv, line 287:
```suggestion
            branches=['3.9', '3.10', '3.11', '3.x'],
```

master/custom/factories.py
```diff
@@ -557,6 +557,15 @@ class Windows64Build(BaseWindowsBuild):
...
+    test_timeout = TEST_TIMEOUT * 2
+    cleanFlags = ["-p", "x64"]
+    factory_tags = ["win64", "bigmem"]
```
ambv, line 566: Is it fine to just add a new tag?

master/custom/workers.py
zware: This should go to the regrtest `-j` option to set the number of parallel test workers explicitly.

master/custom/factories.py
zware: This is for disambiguation in case there are multiple similar builders on a worker.  The working directory is constructed something like `<local worker workdir prefix>/<branchname>.<workername><buildersuffix>`.  For example, on my Gentoo bot I have `/buildbot/buildarea/cpython/3.x.ware-gentoo-x86.nondebug/` and `/buildbot/buildarea/cpython/3.x.ware-gentoo-x86.installed/`.

master/custom/factories.py
zware: Yes, it's a duplicate.  It should probably be set by the worker rather than the builder (as in, leave out `-j` here), in case another worker does this build and wants a different parallelization value.

master/custom/workers.py
zware: Actually, come to think, you can leave this out.  The default is all current branches.

master/custom/factories.py
zware: Yep

PR #358, opened 2022-12-14
Implement comment-based commands for triggering buildbots and new labels

pablogsal:
This commit implements the following:

- A new comment-based way to trigger specific buildbots based on regular
  expressions that will be tested against builder names.
- A new label to trigger specifically refleak builders.

### DIFF
--- a/master/custom/pr_testing.py
+++ b/master/custom/pr_testing.py
@@ -1,4 +1,5 @@
 # Functions to enable buildbot to test pull requests and report back
+import re
 import logging
 
 from dateutil.parser import parse as dateparse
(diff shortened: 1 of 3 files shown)


master/custom/pr_testing.py
```diff
@@ -111,5 +175,143 @@ def handle_pull_request(self, payload, event):
...
+                "GitHub PR #{}, Ignoring: "
+                "head commit message contains skip pattern".format(number)
+            )
+            return ([], "git")
+
+        yield self._remove_label_and_comment(payload, label)
+
+        builder_filter = ""
+        if label == TESTING_LABEL:
+            builder_filter = ".*"
```
zware, line 311: Should we try to filter out the refleak builders here?

zware (approved):
Looks reasonable to me, though I'll confess that I haven't taken the time to fully understand all of the details.

master/custom/pr_testing.py
pablogsal: Not really, as the current test label means "test with all buildbots" and I don't want people that has been used to that meaning to realize too late that the label didn't ran the refleak ones.

PR #359, opened 2023-01-27
Add gps-arm64-debian, a Clang LTO + PGO bot

gpshead:


master/custom/workers.py
```diff
@@ -198,6 +198,12 @@ def get_workers(settings):
             name="gps-debian-profile-opt",
             tags=['linux', 'unix', 'debian', 'amd64', 'x86-64'],
         ),
+        cpw(
+            name="gps-arm64-debian",
+            tags=['linux', 'unix', 'arm64', 'aarch64', 'arm', 'debian'],
+            parallel_tests=7,  # Shortest test time; 4 vCPU host.
+            branches=['3.10', '3.11', '3.x'],
```
zware, line 205: Is 3.9 problematic for this host?  If not, we'd be better off leaving out the `branches` specifier.

If it is, we really ought to have a `not_branches` to go along with `branches` :)

zware (commented):
Once merged, you should be able to connect with the same password as your `gps-raspbian` worker.

master/custom/workers.py
gpshead: 3.9 isn't accepting bugfixes anymore. The way i have this bot configured, `test_tarfile` will fail due to the high uid and gid i gave the buildbot user to trigger https://github.com/python/cpython/issues/101334.

I agree that an exclusion list would be easier to maintain. I don't want to rework major logic in this config adding PR.

zware (approved):

PR #361, opened 2023-02-10
Make Bedevere showing label names like GitHub does

arhadthedev:
Currently Bedevere [prints raw label names](https://github.com/python/cpython/pull/101748#issuecomment-1424532890):

Note that a label is spelled as raw `:hammer: test-with-refleak-buildbots`. However, the label selector shows it differently:

This PR fixed the raw output of code blocks with a manual rendering of this exact emoticon. Thanks to it the message will look like this (simulated with Chrome Developer Tools):

### DIFF
--- a/master/custom/pr_testing.py
+++ b/master/custom/pr_testing.py
@@ -18,7 +18,7 @@ GITHUB_PROPERTIES_WHITELIST = ["*.labels"]
 BUILD_SCHEDULED_MESSAGE = f"""\
 :robot: New build scheduled with the buildbot fleet by @{{user}} for commit {{commit}} :robot:
 
-If you want to schedule another build, you need to add the `{{label}}` label again.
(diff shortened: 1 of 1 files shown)


arhadthedev:
@zware

hugovk:
Thanks for this, it was bugging me recently too :)

Rather than using `code formatting`, does it look better with the `<kbd>` tag?

So instead of:

If you want to schedule another build, you need to add the `🔨 test-with-refleak-buildbots` label again.

Like this:



If you want to schedule another build, you need to add the <kbd>:hammer: test-with-refleak-buildbots</kbd> label again.


It works with either of these:

```md
<kbd>:hammer: test-with-refleak-buildbots</kbd>

<kbd>🔨 test-with-refleak-buildbots</kbd>
```

arhadthedev:
@hugovk Thank you, it made the patch extremely compact.

hugovk (approved):

zware (approved):

PR #375, opened 2023-05-27
Remove kloth-win11 and kloth-win64 workers

zware:
They served long and faithfully, but haven't yet returned.

They'll be welcomed back if ever they reappear :)


jkloth:
Sorry, no permanent home for my buildbots as of yet.   I'm sad to see those machines sitting unused, but I do not foresee a usable location anytime soon.

zware:
Thanks for the update.  I'll leave things in place on the master such that they can be put back in service with as little friction as possible if they do find a place :)

Thanks again for running them for so long!

PR #377, opened 2023-07-25
Update requirements

vstinner:


vstinner:
Sadly, this change broke the server :-( https://github.com/python/buildmaster-config/issues/379

PR #386, opened 2023-08-31
Remove koobs-freebsd-9e36 and koobs-freebsd-564d workers

zware:
With many thanks for years of faithful service.

Should they be able to return in the future, they will be welcomed back :)


vstinner:
Oh no, that makes me sad. Python has no FreeBSD buildbot anymore?

As you wrote, thank you **very much** @koobs for maintaining these servers for many years! They were very useful.

koobs:
I'll be back

PR #388, opened 2023-09-05
Add ware-freebsd worker

zware:


vstinner:
Woooot! That's very cool! I started to fix the first FreeBSD failure :-) https://github.com/python/cpython/pull/108964

PR #397, opened 2023-09-14
Fix stop-master

vstinner:


vstinner:
I suck at sysadmin, soooooorry.

PR #400, opened 2023-09-21
Close #390: Ignore !buildbot comment if no builder matches

### LINKED ISSUE #390: "!buildbot regex" runs all buildbots if the regex matches no worker
The ``!buildbot Raspbian`` command didn't match any buildbot worker, but bedevere-bot decided instead to test the PR on **all** buildbot workers in this case: https://github.com/python/cpython/pull/109101#issuecomment-1710691518

The ``!buildbot ARM Raspbian`` command matched ``ARM Raspbian PR`` as expected: https://github.com/python/cpython/pull/109101#issuecomment-1710692439

By the way, it would be nice to match the regex anywork in the worker name, not only at the beginning of the name.

vstinner:
Also, match the regex anywhere in the builder name, not only at the start.

### DIFF
--- a/master/custom/pr_testing.py
+++ b/master/custom/pr_testing.py
@@ -255,6 +255,20 @@ class CustomGitHubEventHandler(GitHubEventHandler):
             return ([], "git")
 
         builder_filter_fn = re.compile(builder_filter)
+        matched_builders = [
+            builder
+            for builder in self.builder_names
+            if builder_filter_fn.search(builder)
+        ]
+        if not matched_builders:
+            log.msg(f"GitHub PR #{number}: regex {builder_filter!r} "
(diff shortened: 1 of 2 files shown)


vstinner:
Ok, the change works as expected: https://github.com/python/cpython/pull/109698#issuecomment-1730474090

I just made a typo, fixed by: https://github.com/python/buildmaster-config/commit/69e9ee1c439089ed989869dbfcf5cb036be3a836

vstinner:
> ‼️ NameError: name 'builder' is not defined custom.pr_testing in <listcomp> [View Issue](https://python-software-foundation.sentry.io/issues/4512663772/?referrer=github-pr-bot)

I fixed it.

PR #402, opened 2023-09-22
Mark FreeBSD Tier-3 buildbot as stable

vstinner:


vstinner:
See: https://discuss.python.org/t/freebsd-gets-a-new-cirrus-ci-github-action-job-and-a-new-buildbot/33122/28

cc @zware

PR #405, opened 2023-09-26
run make regen-requirements

vstinner:


vstinner:
venv was created by Salt, the server was restarted and buildbot is running on Buildbot 3.9.2, nice!

PR #395, opened 2023-09-13
gh-392: Add Windows NoGIL builder & worker

itamaro:
Add an AWS AMD64 Windows buildbot worker
Add a NoGIL Windows build factory
Add a NoGIL Windows builder to the newly added AWS Windows worker

fixes #392

Issue #384, opened 2023-08-25
The buildbot server no longer start new builds


vstinner:
```
$ date
Fri Aug 25 09:51:55 UTC 2023

$ grep 'starting build ' master/twistd.log
# empty

$ grep 'starting build ' master/twistd.log.1|tail

2023-08-24 16:38:47+0000 [-] starting build <Build AMD64 Fedora Stable Clang Installed 3.x number:4304 results:success>.. pinging the worker <WorkerForBuilder builder='AMD64 Fedora Stable Clang Installed 3.x' worker='cstratak-fedora-stable-x86_64' state=BUILDING>
2023-08-24 16:38:47+0000 [-] starting build <Build AMD64 Fedora Stable LTO + PGO 3.x number:4551 results:success>.. pinging the worker <WorkerForBuilder builder='AMD64 Fedora Stable LTO + PGO 3.x' worker='cstratak-fedora-stable-x86_64' state=BUILDING>
2023-08-24 16:45:35+0000 [-] starting build <Build s390x Fedora Rawhide 3.x number:None results:success> using worker <WorkerForBuilder builder='s390x Fedora Rawhide 3.x' worker='edelsohn-fedora-rawhide-z' state=AVAILABLE>
2023-08-24 16:45:35+0000 [-] starting build <Build s390x Fedora Rawhide Clang 3.x number:None results:success> using worker <WorkerForBuilder builder='s390x Fedora Rawhide Clang 3.x' worker='edelsohn-fedora-rawhide-z' state=AVAILABLE>
2023-08-24 16:45:36+0000 [-] starting build <Build s390x Fedora Rawhide Clang Installed 3.x number:None results:success> using worker <WorkerForBuilder builder='s390x Fedora Rawhide Clang Installed 3.x' worker='edelsohn-fedora-rawhide-z' state=AVAILABLE>
2023-08-24 16:45:36+0000 [-] starting build <Build s390x Fedora Rawhide 3.x number:3548 results:success>.. pinging the worker <WorkerForBuilder builder='s390x Fedora Rawhide 3.x' worker='edelsohn-fedora-rawhide-z' state=BUILDING>
2023-08-24 16:45:36+0000 [-] starting build <Build s390x Fedora Rawhide LTO + PGO 3.x number:None results:success> using worker <WorkerForBuilder builder='s390x Fedora Rawhide LTO + PGO 3.x' worker='edelsohn-fedora-rawhide-z' state=AVAILABLE>
[...]

PR #407, opened 2023-09-27
Fix use of deprecated config options

zware:
See buildbot/buildbot#6875

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -496,13 +496,13 @@ c["services"].append(
 # from buildbot.master import Manhole
 # c['manhole'] = Manhole(9999, "admin", "oneddens")
 
-# the 'projectName' string will be used to describe the project that this
+# the 'title' string will be used to describe the project that this
 # buildbot is working on. For example, it is used as the title of the
-# waterfall HTML page. The 'projectURL' string will be used to provide a link
+# waterfall HTML page. The 'titleURL' string will be used to provide a link
 # from buildbot HTML pages to your project's home page.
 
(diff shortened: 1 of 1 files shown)


vstinner:
How did you notice that? Did you see messages in logs?

zware:
Yep, log messages when manually deploying #395.  I opted to manually pull and `reconfig` rather than allow an automatic `restart` to avoid forgetting about it and letting it hang :)

vstinner:
> Yep, log messages when manually deploying https://github.com/python/buildmaster-config/pull/395. I opted to manually pull and reconfig rather than allow an automatic restart to avoid forgetting about it and letting it hang :)

I fixed ``make stop-master`` recently.

```
stop-master: TARGET=stop
stop-master: run-target
        # issue #384: sometimes when "buildbot stop master" sends SIGINT to
        # Twisted, the server goes in a broken state: it's being "shut down",
        # but it never completes. The server stays forever in this state: it is
        # still "running" but no longer schedules new jobs. Kill the process
        # to make sure that it goes bad to a known state (don't run anymore).
        echo "Buildbot processes (look for process name: $(PKILL_NAME))"
        pgrep -u $(USER) $(PKILL_NAME) ||:
        echo "Send SIGKILL to remaining buildbot processes (if any)"
        pkill -KILL -u $(USER) $(PKILL_NAME) ||:
```

Salt deployment should no longer hang the server: https://github.com/python/buildmaster-config/issues/384

If you deploy manually the config, Salt may restart the server one more time. But (I'm not sure about that.)

PR #417, opened 2023-10-04
Fix UnixNoGilRefleakBuild.testFlags

vstinner:
It must be a list, no a string.

Correctly, the builder fails with:

    ./python -E  -m test --slow-ci --timeout=12000 - R   3 : 3   - u - c p u --junit-xml test-results.xml -j2 --fail-rerun
    regrtest.py: error: unrecognized arguments: -


vstinner:
cc @itamaro: I made changes recently, maybe your PR landed "while" I was doing these changes.

itamaro:
Thank you for fixing it!

Issue #419, opened 2023-10-06
Allow triggering unstable buildbots with the !buildbot PR comment


itamaro:
this came up several times recently around testing the NoGIL build on PRs - it's currently not possible to do that using the `!buildbot nogil` comment ([example attempt](https://github.com/python/cpython/pull/110480#issuecomment-1750962617)) because the nogil builders are marked as unstable.
considering the NoGIL build is "experimental", keeping these builders as "unstable" is probably a good idea, but then we need some way to be able to request these builders on PRs.

after chatting with @ambv, I propose that we extend the `!buildbot` command to support unstable builders.
the mechanics can be discussed in this issue. here are the options I am considering:
1. easy but changes existing behavior: include unstable buildbots by default
2. also easy but kinda hacky: include unstable buildbots *if they have a certain tag* (e.g. "trigger-on-pr-command")
3. introduce new spelling for explicitly allowing unstable builders (e.g. `!buildbot_with_unstable ...` or `!buildbot[unstable]` or `!buildbot ... #with_unstable` - bikeshedding welcome)

vstinner:
An alternative is to promote the buildbot to stable.

* https://buildbot.python.org/all/#/builders/1241 : multiple builds succeeded in a row, I would be ok to mark it as STABLE. We can downgrade it to UNSTABLE if it becomes unstable again.
* https://buildbot.python.org/all/#/builders/1225 most builds are success, same.

Do you want to propose a PR to make them STABLE?

itamaro:
I'm fine with changing these to STABLE, and happy to send a PR to make that change!
I was under the impression that marking a builder as STABLE has broader implications that are not appropriate for an experimental and optional build mode (not sure what my impression was based on, other than my own interpretation of the word "stable" :) )

ambv:
You're correct, Itamar. Stable buildbots are special for release managers, so at least until the PEP is officially approved, I'd wait with marking nogil BBs as stable.

I believe that the limitation to only allow running stable buildbots with "!buildbot" is unnecessary.

vstinner:
Oh wait, I was confused about the definition of "stable" here.

For me, unstable means "is known to fail", and stable "is known to be reliable. Don't we have tiers to decide if a builder blocks a released or not? For example, Tier 3 failures do **not** block a release: https://peps.python.org/pep-0011/#tier-3

itamaro:
[...]

PR #420, opened 2023-10-08
gh-419: Allow unstable builders to trigger on PRs

itamaro:
This intention of this PR is to make it possible to *manually trigger* also unstable buildbots on PRs using the `!buildbot` command (and labels).

The lists of "PR builders" are used in several places where I kept it to *only* stable PR builders, because those other places didn't seem related to the manual triggering of buildbots on PRs.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -313,8 +313,7 @@ for name, worker_name, buildfactory, stability, tier in BUILDERS:
     if any(pattern in buildername for pattern in NO_NOTIFICATION):
         continue
 
-    if stability == STABLE:
-        stable_pull_request_builders.append(buildername)
+    stable_pull_request_builders.append(buildername)
 
     source = GitHub(repourl=git_url, **GIT_KWDS)
 
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -306,14 +306,15 @@ for name, worker_name, buildfactory, stability, tier in BUILDERS:
 
     buildername = name + " " + "PR"
 
+    if stability != STABLE:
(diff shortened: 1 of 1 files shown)


master/master.cfg
```diff
@@ -311,7 +309,11 @@ for name, worker_name, buildfactory, stability, tier in BUILDERS:
     if any(pattern in buildername for pattern in NO_NOTIFICATION):
         continue
 
-    pull_request_builders.append(buildername)
+    if stability == STABLE:
+        stable_pull_request_builders.append(buildername)
+    else:
+        unstable_pull_request_builders.append(buildername)
```
vstinner, line 315: You should move this before NO_NOTIFICATION test.

itamaro: done

colesbury:
I would really like to see this change, not only for the nogil buildbots but also so I can test changes on AIX. The AIX buildbots are also marked as "unstable".

itamaro:
@vstinner I am not sure how to test changes like this manually / locally, as it seems highly coupled with the details of the actual deployed buildbot master.

I think we do want to allow using a PR `!buildbot ...` comment to match unstable builders, but using the `test-with-buildbots` or `test-with-refleak-buildbots` labels should trigger only stable builders.

both PR labels and comments are handled in `CustomGitHubEventHandler`, but it appears that the `builder_names` argument is only used when doing matching against a PR comment. in case of a PR label, I don't see how exactly the builder names are plumbed through, and it isn't clear to me whether this change would impact the behavior of the label-based triggering, or only the comment-based triggering.

PR #421, opened 2023-10-09
Replace ware-gentoo-x86 with ware-debian-x86

zware:


vstinner:
Welcome to Debian :-)

PR #422, opened 2023-10-11
Issue #419: Only add Tier-1 and Tier-2 builders to Release Status

vstinner:


vstinner:
cc @ambv @pablogsal @zware

master/master.cfg
```diff
@@ -252,6 +253,9 @@ for branch_num, (git_url, branchname, git_branch) in enumerate(git_branches):
             # disable GitHub notifications for unstable builders
             if stability == STABLE:
                 github_status_builders.append(buildername)
+                # Only Tier-1 and Tier-2 builders can block a release
+                if tier in (TIER_1, TIER_2):
```
ambv, line 257: builtins.NameError: name 'TIER_1' is not defined

vstinner: Fixed ;-)

vstinner:
cc @pablogsal @Yhg1s @ambv: I fixed https://buildbot.python.org/all/#/release_status page to only list buildbot builders of Tier-1 and Tier-2. Builders of Tier-3 or with no Tier must not block a release and so are no longer listed.

PR #423, opened 2023-10-11
Cache Release Status result for 5 minutes

vstinner:


vstinner:
@pablogsal @ambv: Military grade optimization, I reduced the cache to 5 minutes.

master/custom/templates/releasedashboard.html
```diff
@@ -46,4 +46,12 @@ <h4 class="panel-title">
...
         </div>
     </div>
+    <div class="container">
+        <small>Generated at <time id="generatedAt" datetime="{{generated_at}}">{{generated_at}}</time></small>
+    </div>
 </div>
+<script type="javascript">
+    let elem = document.getElementById("generatedAt");
+    let date = new Date(elem.dateTime);
+    elem.innerText = date.toLocaleDateString("en-CA") + " " + date.toLocaleTimeString("pl-PL");
```
vstinner, line 56:
```suggestion
    elem.innerText = date.toLocaleDateString("en-US") + " (UTC)";
```

master/custom/templates/releasedashboard.html
```diff
@@ -46,4 +46,12 @@ <h4 class="panel-title">
             </div>
         </div>
     </div>
+    <div class="container">
+        <small>Generated at <time id="generatedAt" datetime="{{generated_at}}">{{generated_at}}</time></small>
```
vstinner, line 50:
```suggestion
        <small>Generated at <time id="generatedAt" datetime="{{generated_at}}">{{generated_at}}</time> (cached for 5 minutes)</small>
```

vstinner: Maybe the cache duration should be a template parameter.

ambv: No, this won't be cached for 5 minutes.

master/custom/templates/releasedashboard.html
ambv: No, this is invalid.

master/custom/release_dashboard.py
```diff
@@ -53,7 +58,28 @@ def main():
 
             failed_builds_by_branch[branch].append((builder, last_build))
 
+        generated_at = datetime.datetime.now(tz=datetime.UTC)
```
ambv, line 61:
```
datetime.UTC
Alias for the UTC timezone singleton [datetime.timezone.utc](dfile:///Users/ambv/Library/Application%20Support/Dash/DocSets/Python_3/Python%203.docset/Contents/Resources/Documents/doc/library/datetime.html#datetime.timezone.utc).

New in version 3.11.
```

vstinner:
> @ambv: Use datetime.timezone.utc which is available on 3.9

@zware: Once you mentioned that you considered to upgrade Python on the buildbot? Is it still the case? :-)

master/custom/templates/releasedashboard.html
vstinner: Why is the date formatted twice, in 2 different locales?

zware:
> @zware: Once you mentioned that you considered to upgrade Python on the buildbot? Is it still the case? :-)

I'm in favor, but have made no attempts towards it (that I remember :)).  Pretty sure it will need to be done via the [Salt configuration](https://github.com/python/psf-salt/blob/main/salt/buildbot/init.sls).

PR #430, opened 2023-10-27
Update itamaro workers configuration

itamaro:
1. Remove the fedora worker - with the AWS-hosted ubuntu worker, better stop relying on a laptop running in my home-office
2. Increase parallel tests for the ubuntu and Windows workers
3. Add missing branch exclusions to the MacOS worker


vstinner:
Merged

PR #433, opened 2023-10-30
Always create PR builders, but only schedule stable ones on label

zware:


master/master.cfg
ambv, line 1: We need to also include the unstable PR workers in the GitHubPrScheduler, no?

ambv: See line 355 (somehow GitHub refused to put the comment under a line you didn't touch; sorry!).

zware: I don't think so, no.  That would cause them to be triggered by the label, which AFAIK is not what we want.

ambv: Right!

zware: I may have it wrong and that will be #434, we'll see :)

PR #434, opened 2023-10-30
GH-419: Pass unstable builders also into GitHubPrScheduler

itamaro:
This is needed to support triggering unstable builders using `!buildbot` PR comment

Pass `GitHubPrScheduler` a set of "known stable builders" to it can avoid triggering unstable builders unless it's a comment-based trigger

### DIFF
--- a/master/custom/schedulers.py
+++ b/master/custom/schedulers.py
@@ -5,6 +5,10 @@ from twisted.python import log
 
 
 class GitHubPrScheduler(AnyBranchScheduler):
+    def __init__(self, *args, stable_builder_names, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.stable_builder_names = stable_builder_names
+
     @defer.inlineCallbacks
     def addBuildsetForChanges(self, **kwargs):
         log.msg("Preapring buildset for PR changes")
@@ -18,8 +22,16 @@ class GitHubPrScheduler(AnyBranchScheduler):
         chdict = yield self.master.db.changes.getChange(changeid)
 
         builder_filter = chdict["properties"].get("builderfilter", None)
+        event = chdict["properties"].get("event", None)
         builder_names = kwargs.get("builderNames", self.builderNames)
         if builder_filter and builder_names:
+            # allow unstable builders only for comment-based trigger
+            if event != "issue_comment":
+                builder_names = [
+                    builder_name
+                    for builder_name in builder_names
+                    if builder_name in self.stable_builder_names
+                ]
             log.msg("Found builder filter: {}".format(builder_filter))
             builder_filter, _ = builder_filter
             matcher = re.compile(builder_filter, re.IGNORECASE)
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -309,13 +309,12 @@ for name, worker_name, buildfactory, stability, tier in BUILDERS:
     if "Windows7" in name:
         continue
 
-    buildername = name + " " + "PR"
+    buildername = f"{name} {stability} PR"
 
-    if stability != STABLE:
-        unstable_pull_request_builders.append(buildername)
-    else:
+    if stability == STABLE:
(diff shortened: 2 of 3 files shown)


master/master.cfg
```diff
@@ -309,13 +309,12 @@ for name, worker_name, buildfactory, stability, tier in BUILDERS:
...
-        unstable_pull_request_builders.append(buildername)
-    else:
+    if stability == STABLE:
         stable_pull_request_builders.append(buildername)
-
+    else:
+        unstable_pull_request_builders.append(buildername)
 
     if any(pattern in buildername for pattern in NO_NOTIFICATION):
         continue
```
zware, line 320: I think if you axe this check, we'd land about where we want to be.

itamaro: I tried swapping the order of this check with adding the builder name to the list - that passes the config check!

it would work to axe it as well, but that would change the behavior for builders that match the patterns in `NO_NOTIFICATION`, and I don't know if that's a desirable change (I'm assuming there's a reason for that).

itamaro: with either change (swap order or axe it), I think either approach would work - do you have a preference which one to take?

1. add "stability" to the builder name and use it in the regex; or
2. the current state of this PR - use an explicit set of "stable builders" when not doing comment-based triggering

zware: I'd say 2.

I think we can do away with `NO_NOTIFICATION`.  The Alpine and UBSan builders are unstable, so should be left out of label builds anyway, and we no longer have a Cygwin builder.  I could see value in being able to manually trigger the Alpine or UBSan builders via a comment.

zware: Ok, scratch that: we can't quite do away with `NO_NOTIFICATION`, but we can drop the check on it here.

itamaro: ok, I dropped the check, I think this should be good to go!

zware (approved):

master/master.cfg
```diff
@@ -309,16 +309,12 @@ for name, worker_name, buildfactory, stability, tier in BUILDERS:
...
 
-    if stability != STABLE:
-        unstable_pull_request_builders.append(buildername)
-    else:
+    if stability == STABLE:
         stable_pull_request_builders.append(buildername)
-
-
-    if any(pattern in buildername for pattern in NO_NOTIFICATION):
-        continue
```
vstinner, line 321: Removing NO_NOTIFICATION test is correct, since it's also tested at line 249.

master/custom/schedulers.py
```diff
@@ -18,8 +22,16 @@ def addBuildsetForChanges(self, **kwargs):
...
+        event = chdict["properties"].get("event", None)
         builder_names = kwargs.get("builderNames", self.builderNames)
         if builder_filter and builder_names:
+            # allow unstable builders only for comment-based trigger
+            if event != "issue_comment":
+                builder_names = [
+                    builder_name
+                    for builder_name in builder_names
+                    if builder_name in self.stable_builder_names
+                ]
```
vstinner, line 34: Would it be possible to create the 2 build name lists only once in the constructor, rather than having to do it at each GitHub event?

itamaro: not in the general case - you can see in line 26 the initial value of `builder_names` can come from the `kwargs` of the `addBuildsetForChanges` call, so we can't assume anything in the constructor.

I am not an expert on buildbot internals, but I don't actually see where `addBuildsetForChanges` *can* be called with a `builderNames` kwarg. if we only ever use `self.builderNames` in practice, we can definitely pre-construct the list in the constructor.

I can make this change now, or perhaps add logging so we can see whether it is ever called with a `builderNames` kwarg.

vstinner: Oh ok, I was wrong, ignore my comment.

vstinner:
I merged your PR, thanks @itamaro.

PR #435, opened 2023-10-31
run make regen-requirements

vstinner:


vstinner:
Oh, it seems like this update was not enough to fix https://github.com/python/buildmaster-config/security/dependabot/27 security issue.

PR #436, opened 2023-10-31
GH-419: Follow-up fix to support unstable builders triggering

itamaro:
GH-434 attempted to support triggering unstable builder with comment-based triggering It had two flaws that are fixed in this follow-up PR:
1. Needed to unpack the `event` property
2. If we are left with no builders after filtering, need to avoid the super() call (otherwise we end up triggering ALL builders O_O)


vstinner:
Well, I trust you, let's see how it goes :-)

PR #437, opened 2023-11-01
GH-419: Allow unstable builders on GitHub PR Checks

itamaro:
With GH-436 the builders are finally triggered correctly!
But they don't show up in the GitHub checks status thingie. To have these builders show up in GitHub, we need to add them to the `GitHubStatusPush` reporter.


vstinner:
Merged, thanks.

PR #439, opened 2023-11-07
run make regen-requirements

vstinner:


vstinner:
I messed up with Git and it closed my PR... new attempt: PR #440.

PR #440, opened 2023-11-07
run make regen-requirements

vstinner:


vstinner:
> Should fix https://github.com/python/buildmaster-config/security/dependabot/27

Oh. In fact, I'm trying to fix this issue for 2 weeks, but sadly, Twisted is stuck at 23.10.0rc1: release candidate. ``make regen-requirements`` does not install RC versions, and so this PR doesn't fix the security issue.

I tried adding ``--pre`` to ``pip install``, but it didn't help. I leave the issue unfixed for now.

cc @pablogsal @zware @ambv

zware:
Twisted 23.10.0 is out, but `buildbot` has `Twisted` pinned to `<=22.10.0`: https://github.com/buildbot/buildbot/blob/09a54abffc1ea98dae2fe39df058759c2930a118/master/setup.py#L470-L477

vstinner:
Oh! Maybe we can just allow more recent Twisted versions. Is there a reason to block upgrade?

zware:
Apparently buildbot 3.9.1 did not pin Twisted, but I'm not sure how well downgrading from 3.9.2 to 3.9.1 would go.  See buildbot/buildbot#7073 for more context.  Ideally buildbot 3.9.3 will be released Soon™ without the pin.

vstinner:
Sorry, I was even more confused. So buildbot is blocking upgrading to a more recent Twisted version, not https://github.com/python/buildmaster-config/ project. So Buildbot (setup.py) prevents installing the security fix, ok.

zware:
Correct.

vstinner:
I created https://github.com/buildbot/buildbot/issues/7217 to ask buildbot to relax their dependency version on Twisted.

PR #444, opened 2023-11-18
Mark NoGIL Linux & MacOS builders as stable

itamaro:
these builders have been very stable over the last few weeks.
with the [official acceptance of PEP 703](https://discuss.python.org/t/pep-703-making-the-global-interpreter-lock-optional-in-cpython-acceptance/37075), I propose we promote these builders to a stable status.

note that the M1 builders was previously wrongly marked as "tier-1 unstable". according to [PEP 11](https://peps.python.org/pep-0011/), Apple aarch64 is still considered tier-2, so until that changes, I moved the M1 NoGIL builder to "tier-2 stable".

### DIFF
--- a/master/custom/builders.py
+++ b/master/custom/builders.py
@@ -80,6 +80,8 @@ STABLE_BUILDERS_TIER_1 = [
     ("AMD64 RHEL8 Refleaks", "cstratak-RHEL8-x86_64", UnixRefleakBuild),
     ("AMD64 RHEL8 LTO", "cstratak-RHEL8-x86_64", LTONonDebugUnixBuild),
     ("AMD64 RHEL8 LTO + PGO", "cstratak-RHEL8-x86_64", LTOPGONonDebugBuild),
+    ("AMD64 Ubuntu NoGIL", "itamaro-ubuntu-aws", UnixNoGilBuild),
+    ("AMD64 Ubuntu NoGIL Refleaks", "itamaro-ubuntu-aws", UnixNoGilRefleakBuild),
 
     # Windows x86-64 MSVC
     ("AMD64 Windows10", "bolen-windows10", Windows64Build),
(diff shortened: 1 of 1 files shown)


master/custom/builders.py
```diff
@@ -89,6 +91,7 @@
 
     # macOS x86-64 clang
     ("x86-64 macOS", "billenstein-macos", UnixBuild),
+    ("x86-64 MacOS Intel NoGIL", "itamaro-macos-intel-aws", UnixNoGilBuild),
```
corona10, line 94: If free-threaded CIs are triggered anyway, Can we update this buildbot as refleak checker?

itamaro: afaik the buildbot test suite is more comprehensive than the equivalent CI test suite (e.g. due to `--fast-ci` flag for example), so there's value in a buildbot builder with "equivalent configuration". I also believe the release manager considers the buildbot statuses when making release decisions, not GitHub Actions.

I also noticed there are no refleak builders for MacOS at all (for the default build) -- would it be valuable to add such builders? (for default and/or nogil)

corona10: > afaik the buildbot test suite is more comprehensive than the equivalent CI test suite

I see!

> I also noticed there are no refleak builders for MacOS at all (for the default build) -- would it be valuable to add such builders? (for default and/or nogil)

Oh, I didn't know that :( If possible for me, "yes"

itamaro: > If possible for me, "yes"

let me keep the existing builders as is (just marked as stable), and I'll look into adding refleak macos builders separately (probably marked "unstable" initially)

hugovk: > I also believe the release manager considers the buildbot statuses when making release decisions, not GitHub Actions.

Yes, it's here: https://peps.python.org/pep-0101/#how-to-make-a-release

> * Check the stable buildbots.
> 
>   Go to https://buildbot.python.org/all/#/release_status
>
>   Look at the buildbots for the release you’re making. Ignore any that are offline (or inform the community so they can be restarted). If what remains are (mostly) green buildbots, you’re good to go. If you have non-offline red buildbots, you may want to hold up the release until they are fixed. Review the problems and use your judgement, taking into account whether you are making an alpha, beta, or final release.

Although GitHub Actions is kind of implicitly included, because all PRs should be green before merging in the first place.

PR #453, opened 2024-01-16
run make regen-requirements

vstinner:


vstinner:
I'm tired of getting emails from the Twisted vulnerability: https://github.com/python/buildmaster-config/security/dependabot/27

buildbot got updated recently to avoid an old dependency which prevented to upgrade to recent Twisted: https://github.com/python/buildmaster-config/security/dependabot/27

So let's upgrade everything :-)

PR #458, opened 2024-01-22
Add `--disable-gil` test flag to Windows NoGIL builder

itamaro:
since https://github.com/python/cpython/pull/113129 (part of https://github.com/python/cpython/issues/112984), we need to pass `--disable-gil` to `rt.bat` too

### DIFF
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -626,7 +626,8 @@ class Windows64ReleaseBuild(Windows64Build):
 
 class Windows64NoGilBuild(Windows64Build):
     buildersuffix = '.x64.nogil'
-    buildFlags = ["-p", "x64", "--disable-gil"]
+    buildFlags = Windows64Build.buildFlags + ["--disable-gil"]
+    testFlags = Windows64Build.testFlags + ["--disable-gil"]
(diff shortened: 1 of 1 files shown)


zooba:
Not sure if this passes it to `rt.bat` or `Tools\buildbot\test.bat`, but if it's the latter, we'll need to update the test.bat script to pass it through.

itamaro:
> Not sure if this passes it to `rt.bat` or `Tools\buildbot\test.bat`, but if it's the latter, we'll need to update the test.bat script to pass it through.

based on my limited understanding of batch scripts, [this](https://github.com/python/cpython/blob/1d7bddd9612bcbaaedbc837e2936de773e855411/Tools/buildbot/test.bat#L20) would add the `--disable-gil` flag to `regrtest_args`, which would then be handed off to `rt.bat` [here](https://github.com/python/cpython/blob/1d7bddd9612bcbaaedbc837e2936de773e855411/Tools/buildbot/test.bat#L26).

it does look like [recent builds](https://buildbot.python.org/all/#/builders/1241) are now failing in a different way (many `link.exe` failures from `test_peg_generator`).

PR #463, opened 2024-02-06
run make regen-requirements

vstinner:
Update to fix https://github.com/python/buildmaster-config/security/dependabot/31


vstinner:
This change upgrades buildbot from 3.10 to 3.11. I failed to find the changes: https://github.com/buildbot/buildbot/issues/7429

PR #468, opened 2024-02-29
Increase test_timeout on RHEL8-ppc64le for older branches

encukou:
This particular worker is slow: a successful test_gdb run usually will take about 17 minutes, but often it fails on the 20-minute timeout.

On 3.11+, the tests are split up into three 7-min suites so this isn't a problem, but for older branches it's probably best to solve it in the buildbot config.

This is my first foray into buildbot configuration; please tell me if you'd do this differently.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -163,8 +163,8 @@ extra_factory_args = {
         "test_with_PTY": True,
     },
     "cstratak-RHEL8-ppc64le": {
-        # Increase the timeout on specific branches
-        "slow_branches": ['3.8', '3.9', '3.10'],
+        # Increase the timeout on this slow worker
+        "timeout_factor": 2,
     },
 
 }
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -162,6 +162,11 @@ extra_factory_args = {
         # test curses as well
         "test_with_PTY": True,
     },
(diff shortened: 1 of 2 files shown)


master/master.cfg
```diff
@@ -162,6 +162,11 @@ extra_factory_args = {
         # test curses as well
         "test_with_PTY": True,
     },
+    "cstratak-RHEL8-ppc64le": {
+        # Increase the timeout on specific branches
+        "slow_branches": ['3.8', '3.9', '3.10'],
```
vstinner, line 167: It would be simpler to increase the timeout on all branches, no?

encukou: It would, but do we want that?

vstinner: I'm not sure why, but this specific worker is slower than other worker. It got a few timeout issues. I suggest to increase the timeout. I don't see any downside. It's just a hard limit.

encukou: OK.
I didn't find a pre-existing setting for that either, so I added one the same way as `slow_branches` before.
Does this look good?

vstinner (approved):
LGTM

encukou:
Thanks!
I plan to merge this after next week's release.

PR #471, opened 2024-03-05
Remove buildbot_profiler plugin

vstinner:
See: https://github.com/tardyp/buildbot_profiler/issues/7


vstinner:
@pablogsal @ambv: Are you ok with removing the profiler? Or do you still need it?

pablogsal:
> @pablogsal @ambv: Are you ok with removing the profiler? Or do you still need it?

Let's remove it 👍

zware (approved):
Might as well generate requirements, but if you want that to be a separate commit that's fine too.

PR #472, opened 2024-03-05
run make regen-requirements

vstinner:


vstinner:
I deployed the new config manually with `sudo salt-call state.highstate`, it was applied successfully: https://buildbot.python.org/all/#/builders is still running :-)

PR #476, opened 2024-04-11
Increase UnixNoGilBuild timeout

vstinner:


vstinner:
Example: on [build 1985](https://buildbot.python.org/all/#/builders/1225/builds/1985), test_math failed with a timeout (20 min)

```
test_sumprod_stress (test.test_math.MathTests.test_sumprod_stress) ... Timeout (0:20:00)!
```

The [next build](https://buildbot.python.org/all/#/builders/1225/builds/1986) is a success, but test_math was the slowest test: `test_math: 9 min 35 sec`.

test_math bad performance is a known issue, unrelated to Free Threading: https://github.com/python/cpython/issues/110327

vstinner:
cc @itamaro

master/custom/factories.py
```diff
@@ -169,6 +169,9 @@ class UnixNoGilBuild(UnixBuild):
     buildersuffix = ".nogil"
     configureFlags = ["--with-pydebug", "--disable-gil"]
     factory_tags = ["nogil"]
+    # 2024-04-11: Free-threading can still be slower than regular build in some
+    # code paths, so tolerate longer timeout.
+    test_timeout = int(TEST_TIMEOUT * 1.5)
 
 
 class UnixNoGilRefleakBuild(UnixBuild):
```
itamaro, line 177: Should we increase the timeout for this one too?

itamaro (approved):
Thanks Victor!

master/custom/factories.py
vstinner: UnixNoGilRefleakBuild already uses `TEST_TIMEOUT * 10` and I didn't see any test failing with this timeout. I prefer to not make it bigger (until a test fails with timeout), the timeout is already larger than 3 hours! (`worker timeout: 3 hour 25 min`)

vstinner:
Merged, thanks for the review.

Issue #473, opened 2024-03-08
New worker request: <diegorusso-aarch64-bigmem>


diegorusso:
### Username

buildbot

### Name

Diego Russo

### Email address

diego.russo@arm.com

### Password status

I need a new owner password

### Processor architecture

aarch64

### Operating System

Ubuntu 22.04

### Anything special about the worker?

This machine is an aarch64 Equinix machine and it is currently used to run benchmarks to be uploaded to speed.python.org.
Given its spec, we'd like to run also bigmem tests.
Tagging @ambv for visibility

zware:
Hmm, if it's a benchmark runner, do we really want it running a buildbot as well?  It seems that would impact its ability to run benchmarks effectively, or would require routinely stopping the buildbot worker process.

diegorusso:
We have set up the machine with cpu isolation. The machine has 80 cores and we have isolated the last 14 cores for running benchmarks: to be precise we run benchmarks just on the odd cores in the range 65-79 (65, 67, 69..) using cpu affinity.

Daily benchmarks are run at midnight (UTC) only, the rest of the time the machine is idle.

The aim is to maximise the use of this machine. if we want to be on the safe side we could pause buildbot jobs 2 hours before and after midnight so it won't be scheduling jobs when benchmarks will be running.

zware:
Ok.  I'm no expert at benchmarks, so if that configuration is deemed sufficient from that side of things, I'm ok from the buildbot side of things :)

Can you create a PR adding the public configuration?  #404 should be a reasonable template to follow, though there may be some extras needed to enable bigmem testing.

itamaro:
@diegorusso I created a new owner password and emailed it to you.

vstinner:
Related PR: https://github.com/python/buildmaster-config/pull/474

diegorusso:
Thanks @itamaro. I confirmed I received it.

diegorusso:
I have my hands blocked until this https://discuss.python.org/t/top-level-license-not-present-on-some-repos/47973 gets resolved.

itamaro:
@ambv were you planning to add license and CLA bot?

ambv:
MIT license added as all contributors were reached for acknowledgement and provided it.

hugovk:
Can this be closed now https://github.com/python/buildmaster-config/pull/546 is merged and https://buildbot.python.org/?#/workers/114 is active?

diegorusso:
[...]

PR #484, opened 2024-04-21
Ensure a clean copy of configure arguments is used

freakboy3742:
This is a variation on the change introduced by #473.

`host_configure_args` is defined as a class-level variable; however, there are build steps that assign this class-level instance directly to `configure_cmd`, and then concatenate build-specific commands. As a result, configure command line arguments can be duplicated (sometimes in contradictory ways).

I noticed this on the iOS builds; see L3 defining the `args` in [this build](https://buildbot.python.org/all/#/builders/1380/builds/43/steps/6/logs/stdio) - there are essentially three copies of the same command line arguments (`--with-pydebug` et al is defined 3 times).

However, it also impacts on any build that uses the `UnixCrossBuild` base class. AFAICT, the only example of a builder affected by this that is currently in use is the (now deprecated) WASI32 builders; those end up with artefacts like the `--with-pydebug --without-pydebug` arguments in [this build log](https://buildbot.python.org/all/#/builders/1374/builds/871/steps/8/logs/stdio). This doesn't appear to be causing problems at present, but it is at the very least confusing.

This PR takes a copy of the class-level `host_configure_args` variable before modifying it, which should prevent different instance setups tainting each other. This is a pattern that is also followed by the handling of other args (e.g., `defaultTestOpts` and `testFlags`).

### DIFF
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -750,7 +750,9 @@ class UnixCrossBuild(UnixBuild):
 
         # Now that we have a "build" architecture Python, we can use that
         # to build a "host" (also known as the target we are cross compiling)
-        configure_cmd = self.host_configure_cmd + ["--prefix", "$(PWD)/target/host"]
+        # Take a copy so that the class-level definition isn't tainted
+        configure_cmd = list(self.host_configure_cmd)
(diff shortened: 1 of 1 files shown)


vstinner:
LGTM, merged.

PR #489, opened 2024-06-04
Increase Timeouts For the Stable WASI Worker

ericsnowcurrently:
This should resolve https://github.com/python/cpython/issues/118916.


vstinner:
Merged. Let's see if it's enough to repair the WASI worker :-)

PR #490, opened 2024-06-04
Increase parallel tests on itamaro ubuntu worker

colesbury:
The worker has been swapped out for a larger instance type, so we can run more tests in parallel again.


colesbury:
cc @itamaro

PR #494, opened 2024-06-20
Remove buildbot workers hosted on my ex-employer's cloud.

gpshead:
I no longer have access to these GCE VMs.

Something happened on or after 8-9 days ago on or after 939c201e00943c6dc2d515185168c30606ae522c (probably not that change, but I didn't try to figure out what) with the gps-debian-profile-opt bot that likely lead to runaway busy-looping processes for several builds on both 3.x and 3.13 branches, causing it to take 3 hours instead of 20 minutes to complete a run after that.

Time to shut the bot I cannot administer anymore down, lets just take care of all three of them at once.

@Yhg1s still works there and may have access to these bots via the cloud.google.com console on his corp account if I recall correctly who I added to the config. So if you want these to stay up during the 3.13beta releases, please investigate and confirm your access.

If anyone with access at Google wants to take these over and keep them running (very infrequent baby sitting - update & reboot a few times a year, so little I never bothered automating it), they need to reconfigure them and get their own CPython buildbot password to store in the config; follow the instructions linked to at the end of https://www.python.org/dev/buildbot/ to do that.  I don't want my name associated with these bots any longer if they continue to exist.  It is likely easier for you to setup your own replacements than to adopt mine and reconfigure.

### DIFF
--- a/master/custom/builders.py
+++ b/master/custom/builders.py
@@ -69,7 +69,6 @@ NO_TIER = None
 STABLE_BUILDERS_TIER_1 = [
     # Linux x86-64 GCC
     ("AMD64 Debian root", "angelico-debian-amd64", UnixBuild),
-    ("AMD64 Debian PGO", "gps-debian-profile-opt", PGOUnixBuild),
     ("AMD64 Ubuntu Shared", "bolen-ubuntu", SharedUnixBuild),
     ("AMD64 Fedora Stable", "cstratak-fedora-stable-x86_64", FedoraStableBuild),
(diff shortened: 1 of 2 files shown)


vstinner:
In short, this change should not be merged until @Yhg1s can verify if he has access to these machines.

Yhg1s:
I'm not sure why my having access matters, we can just remove these buildbots. UBSan is useful but ASan covers the most important parts of it. I don't think these buildbots provide enough unique coverage that we should go out of our way to try and keep them running for 3.13.

PR #495, opened 2024-07-01
Disable JUnit XML output

vstinner:


vstinner:
@pablogsal @ambv: I chose to keep the code to be able to reenable quickly upload, if someone needs it tomorrow. If you prefer, I can just remove the code.

encukou:
Is there a different way to get machine-readable info on which test failed?

vstinner:
> Is there a different way to get machine-readable info on which test failed?

I don't think so. I propose to disable the XML output since nobody uses it. Except if you use it?

encukou:
Well, I have a semi-working ad-hoc tool, which I haven't used it in a while but has been helpful in the past. I don't know if that counts as using the XML.

But I'm afraid that if I need it again, I'll need it for past dumps, so reenabling the uploads won't help much.

vstinner:
@encukou: If you may still use the XML output, let's keep it for now.

PR #496, opened 2024-07-02
Remove the explicit iOS minimum version.

freakboy3742:
python/cpython#121250 increases the default minimum iOS version to 13.0; however, the buildbot has an explicitly encoded iOS minimum version of 12.0. 

This modifies the builder configuration to fall back to the default unless specifically configured for a particular version.


vstinner:
Merged, thanks.

PR #505, opened 2024-07-17
Keep logs for 6 months

zware:
Also remove the obsolete 'changeHorizon' setting.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -107,13 +107,14 @@ c = BuildmasterConfig = {}
 
 c["db_url"] = str(settings.db_url)
 
-# horizons
-c["changeHorizon"] = 300
-
-# configure a janitor which will delete all logs older than one month,
+# configure a janitor which will delete all logs older than 6 months,
 # and will run on sundays at noon
 c["configurators"] = [
-    util.JanitorConfigurator(logHorizon=timedelta(weeks=4), hour=12, dayOfWeek=6)
+    util.JanitorConfigurator(
+        logHorizon=timedelta(days=180),
+        hour=12,
+        dayOfWeek=6,
+    )
 ]
 
 # workers are set up in workers.py


master/master.cfg
```diff
@@ -107,13 +107,14 @@ c = BuildmasterConfig = {}
...
-# horizons
-c["changeHorizon"] = 300
-
-# configure a janitor which will delete all logs older than one month,
+# configure a janitor which will delete all logs older than 6 months,
 # and will run on sundays at noon
 c["configurators"] = [
-    util.JanitorConfigurator(logHorizon=timedelta(weeks=4), hour=12, dayOfWeek=6)
+    util.JanitorConfigurator(
+        logHorizon=timedelta(days=180),
```
vstinner, line 114: Oh, so currently logs are kept for 30 days. I don't think that storing logs for 180 days is useful. Maybe 2 months, 60 days, is enough.

master/master.cfg
```diff
@@ -107,13 +107,14 @@ c = BuildmasterConfig = {}
 
 c["db_url"] = str(settings.db_url)
 
-# horizons
-c["changeHorizon"] = 300
```
vstinner, line 111: What is this setting?

zware: As far as I can tell, it hasn't actually done anything since the introduction of the `JanitorConfigurator` below.  I think it was supposed to configure how many source revisions to keep data for.

master/master.cfg
zware: As seen on Discord, the leading vote for keeping builds is 6 months, and I'm not sure how useful it is to keep builds without logs.  Unfortunately, logs are currently the only things that can be cleaned up automatically (as far as I can tell).

I think I'd like to at least start with 6 months here and see how much space it takes up.  If 6 months is too much, we can bump it down later.

vstinner (approved):
LGTM.

Well, let's see how it goes :-)

Issue #511, opened 2024-08-15
New worker request: mhsmith-android-arm64


mhsmith:
### Username

mhsmith

### GitHub username

@mhsmith

### Name

Malcolm Smith

### Email address

smith@chaquo.com

### Password status

I need a new owner password

### Processor architecture

ARM64

### Operating System

macOS hosting an Android emulator

### Anything special about the worker?

Required for Android to reach Tier 3 status (PEP 738). I will create a configuration PR in the next few days.

zware:
I've sent you an email with the password that you can use.  We'll need to have merged the configuration before it's useful to you, though :)

itamaro:
Looks like this is done now

PR #516, opened 2024-08-18
Add Android configuration

mhsmith:
* Issue: https://github.com/python/buildmaster-config/issues/511

This PR adds configuration for two Android builders and two workers, one each for ARM64 and x86-64.

Before merging this PR, please also update the settings as follows:
* Rename the worker `mhsmith-android-arm64` to `mhsmith-android-aarch64`, for consistency with the existing workers.
* Add a new worker called `mhsmith-android-x86_64`, with the same password.


zware (approved):

mhsmith:
Thanks!

PR #521, opened 2024-09-13
Update Android configuration

mhsmith:
Changes:

* Add the `-v` option to capture non-Python log messages. These messages have been critical for understanding test failures in the past, including SELinux errors, low memory warnings, and signals being delivered to the wrong thread.

* Since the `-v` option also displays output from building the testbed, there's no longer any need for a separate "Build testbed" step.

* Add the `--single-process --rerun` options to give flaky tests a second chance. This matches the set of options used by the iOS buildbot (cpython/iOS/testbed/iOSTestbedTests/iOSTestbedTests.m).


vstinner (approved):
LGTM

PR #525, opened 2024-09-26
Show all stable builders on the release status page; add Disconnected stable workers

encukou:
Right now the page shows only Tier-1 and Tier-2 builders, and also skips NO_NOTIFICATION builders.

All NO_NOTIFICATION builders are unstable, so this gets rid of NO_NOTIFICATION altogether. Let's use the stable tag instead.

Remove the tier filter.

Add a new section with disconnected workers.

### DIFF
--- a/master/custom/templates/releasedashboard.html
+++ b/master/custom/templates/releasedashboard.html
@@ -32,7 +32,7 @@
                                 <div class="panel-heading">
                                     <div style="text-align: center;">
                                         <h4 class="panel-title">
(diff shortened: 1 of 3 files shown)


master/custom/templates/releasedashboard.html
```diff
@@ -40,7 +42,7 @@ <h4 class="panel-title">
                                         <div>
```
zware, line 42: Should we replace the "Releaseable"/"Not Releaseable" tags with something else?

zware (approved):

PR #519, opened 2024-08-26
Arch Linux Usan 3.x/ build bot with -fno-sanitize=function

pablogsal:
Closes #517

PR #536, opened 2024-10-03
Mark the passing UBSAN build as stable, the failing one as unstable

encukou:
PR #519 added the "AMD64 Arch Linux Usan Function" UBSAN builder with -fno-sanitize=function, but made it unstable and left the existing USAN builder with all checks as stable.

We currently need this skip in order to pass, so the Function one should be marked stable (until gh-111178 is fixed & we can merge the two builders again).


encukou:
I plan to merge after the 3.13.0 release, to avoid disruptions.

encukou:
Thanks `@ambv` for adding me!

PR #539, opened 2024-10-08
Add RHEL8 and RHEL9 s390x workers and builders

stratakis:
Replacing two of the s390x workers that were retired recently.


vstinner:
I sent you passwords by email for your 2 new s390x buildbot workers. Tell me when they are configured with the password, so I can merge this PR.

PR #543, opened 2024-10-17
Reduce refleak timeout from 3h20 to 45min

vstinner:


vstinner:
AMD64 Fedora Stable Refleaks 3.13 slowest test was 9 minutes, whereas a test was blocked for 3 hour 20 min.

cc @zware

vstinner:
* AMD64 Windows11 Refleaks 3.x slowest test: 17 min 11 sec
* ARM64 MacOS M1 Refleaks NoGIL 3.x slowest test: 6 min 14 sec
* s390x RHEL9 Refleaks 3.x: 4 min 48 sec

encukou:
There are slower builders on 3.10 which has monolithic `test_asyncio`. [AMD64 Fedora Stable Refleaks 3.10](https://buildbot.python.org/#/builders/679) and [aarch64 RHEL8 Refleaks 3.10](https://buildbot.python.org/#/builders/639) used to take closer to an hour, and are failing now.
I expect 3.9 builders to start failing when they first build with this config.

Is there any extra background info for this change? Would it be better to revert it, or special-case the branches, like this for example?
```diff
diff --git a/master/custom/factories.py b/master/custom/factories.py
index d4e703e..23dd122 100644
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -72,6 +72,11 @@ class UnixBuild(BaseBuild):
         # Adjust the timeout for this worker
         self.test_timeout *= kwargs.get("timeout_factor", 1)
 
+        # Before 3.11, test_asyncio wasn't split up, so refleaks tests
+        # need more time.
+        if branch in ("3.9", "3.10") and has_option("-R", self.testFlags):
+            self.test_timeout *= 2
+
         if self.build_out_of_tree:
             self.addStep(
                 ShellCommand(
```

zware:
I'm in favor of either that kind of special case, or just dealing with it on the rare occasion that the 3.10 builders run.  I definitely do not want to revert :)

encukou:
I'd like to schedule some extra runs on those builders from time to time. Normal runs are meant for security fixes, and that's a bad time to discover latent issues.

I filed the patch as https://github.com/python/buildmaster-config/pull/545

PR #545, opened 2024-10-24
Increase timeout for 3.9 and 3.10 Refleaks runs

encukou:
These version have monolithic `test_asyncio` and some builders are hitting the 45min limit.

Fix-up for #543.


encukou:
`checkconfig` works.

encukou:
I'll merge tomorrow if there are no objections.

vstinner:
Thanks for the fix.

PR #546, opened 2024-10-24
Add diegorusso-aarch64-bigmem worker

diegorusso:
This worker avoids builds to be started between 10pm and 2am. They will be scheduled at 2am.

This addresses issue https://github.com/python/buildmaster-config/issues/473

### DIFF
--- a/master/custom/builders.py
+++ b/master/custom/builders.py
@@ -8,6 +8,7 @@ from custom.factories import (
     FedoraRawhideFreedthreadingBuild,
     UnixAsanBuild,
     UnixAsanDebugBuild,
+    UnixBigmemBuild,
     UnixTraceRefsBuild,
     UnixVintageParserBuild,
     UnixRefleakBuild,
@@ -124,6 +125,8 @@ STABLE_BUILDERS_TIER_2 = [
     ("aarch64 RHEL8 Refleaks", "cstratak-RHEL8-aarch64", UnixRefleakBuild),
     ("aarch64 RHEL8 LTO", "cstratak-RHEL8-aarch64", LTONonDebugUnixBuild),
     ("aarch64 RHEL8 LTO + PGO", "cstratak-RHEL8-aarch64", LTOPGONonDebugBuild),
+
+    ("aarch64 Ubuntu 22.04 BigMem", "diegorusso-aarch64-bigmem", UnixBigmemBuild),
 
     # macOS aarch64 clang
     ("ARM64 macOS", "pablogsal-macos-m1", MacOSArmWithBrewBuild),
--- a/master/custom/builders.py
+++ b/master/custom/builders.py
@@ -126,8 +126,6 @@ STABLE_BUILDERS_TIER_2 = [
     ("aarch64 RHEL8 LTO", "cstratak-RHEL8-aarch64", LTONonDebugUnixBuild),
     ("aarch64 RHEL8 LTO + PGO", "cstratak-RHEL8-aarch64", LTOPGONonDebugBuild),
 
-    ("aarch64 Ubuntu 22.04 BigMem", "diegorusso-aarch64-bigmem", UnixBigmemBuild),
-
     # macOS aarch64 clang
     ("ARM64 macOS", "pablogsal-macos-m1", MacOSArmWithBrewBuild),
     ("ARM64 MacOS M1 NoGIL", "itamaro-macos-arm64-aws", MacOSArmWithBrewNoGilBuild),
@@ -243,6 +241,8 @@ UNSTABLE_BUILDERS_TIER_2 = [
     ("PPC64LE CentOS9 LTO + PGO", "cstratak-CentOS9-ppc64le", LTOPGONonDebugBuild),
 
     # Linux aarch64 GCC/Clang
+    ("aarch64 Ubuntu 22.04 BigMem", "diegorusso-aarch64-bigmem", UnixBigmemBuild),
+
     # Fedora Rawhide is unstable
     ("aarch64 Fedora Rawhide", "cstratak-fedora-rawhide-aarch64", FedoraRawhideBuild),
     ("aarch64 Fedora Rawhide Refleaks", "cstratak-fedora-rawhide-aarch64", UnixRefleakBuild),
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -267,7 +267,7 @@ class UnixBuildWithoutDocStrings(UnixBuild):
 class UnixBigmemBuild(UnixBuild):
     buildersuffix = ".bigmem"
     testFlags = ["-M60g", "-j4", "-uall,extralargefile"]
-    factory_tags = ["aarch64", "bigmem"]
+    factory_tags = ["bigmem"]
 
 
 class AIXBuild(UnixBuild):
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -264,6 +264,12 @@ class UnixBuildWithoutDocStrings(UnixBuild):
     configureFlags = ["--with-pydebug", "--without-doc-strings"]
 
 
+class UnixBigmemBuild(UnixBuild):
+    buildersuffix = ".bigmem"
+    testFlags = ["-M60g", "-j4", "-uall,extralargefile"]
+    factory_tags = ["aarch64", "bigmem"]
+
+
 class AIXBuild(UnixBuild):
     configureFlags = [
         "--with-pydebug",
--- a/master/custom/workers.py
+++ b/master/custom/workers.py
@@ -148,7 +148,7 @@ def get_workers(settings):
             name="diegorusso-aarch64-bigmem",
             tags=['linux', 'unix', 'ubuntu', 'arm', 'arm64', 'aarch64', 'bigmem'],
             not_branches=['3.9', '3.10', '3.11', '3.12', '3.13'],
-            parallel_tests=60,
+            parallel_tests=4,
         ),
         cpw(
             name="cstratak-rhel8-s390x",
--- a/master/custom/workers.py
+++ b/master/custom/workers.py
@@ -145,6 +145,12 @@ def get_workers(settings):
             parallel_tests=40,
         ),
         cpw(
+            name="diegorusso-aarch64-bigmem",
+            tags=['linux', 'unix', 'ubuntu', 'arm', 'arm64', 'aarch64', 'bigmem'],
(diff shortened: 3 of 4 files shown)


diegorusso:
Folks, I'm not an expert on buildbot. With this worker we want to avoid builds between 10pm and 2am because at midnight we run pyperformance to push data to speed.python.org.
The worker is set at UTC but I'm not sure what's the timezone of the buildbot master. Can anyone shed some lights? if the timezone is not UTC, we need to tweak the time interval.

diegorusso:
Also this worker has:
* Cores: 80 *but* the latest 16 CPUs (ID 64-79) are for running pyperformance and we want to make sure that pyperformance runs with affinity on one of these CPUs. Hence usable cores are 0-63.
* RAM: 250GB

master/custom/builders.py
```diff
@@ -125,6 +126,8 @@
     ("aarch64 RHEL8 LTO", "cstratak-RHEL8-aarch64", LTONonDebugUnixBuild),
     ("aarch64 RHEL8 LTO + PGO", "cstratak-RHEL8-aarch64", LTOPGONonDebugBuild),
 
+    ("aarch64 Ubuntu 22.04 BigMem", "diegorusso-aarch64-bigmem", UnixBigmemBuild),
```
zware, line 129: We need to start in the unstable set, around line 245.  Once we've proven that things work as expected, we can move back here to the stable set.

master/custom/workers.py
```diff
@@ -144,6 +144,12 @@ def get_workers(settings):
             tags=['linux', 'unix', 'rhel', 'arm', 'arm64', 'aarch64'],
             parallel_tests=40,
         ),
+        cpw(
+            name="diegorusso-aarch64-bigmem",
+            tags=['linux', 'unix', 'ubuntu', 'arm', 'arm64', 'aarch64', 'bigmem'],
+            not_branches=['3.9', '3.10', '3.11', '3.12', '3.13'],
```
picnixz, line 150: I'm not sure but can't we use `branches=[MAIN_BRANCH_NAME]`?

master/master.cfg
```diff
@@ -14,7 +14,7 @@ import os
 import subprocess
 import sys
 
-from datetime import timedelta
+from datetime import timedelta, datetime
```
picnixz, line 17:
```suggestion
from datetime import datetime, timedelta
```

master/master.cfg
```diff
@@ -288,6 +288,24 @@ for branch_num, (git_url, branchname, git_branch) in enumerate(git_branches):
...
+def no_builds_between(start, end):
+    def f(builder, requests):
+        now = datetime.now()
+        if start <= now.hour < end:
+            # Calculate the delay until end time
+            delay = ((end - now.hour) % 24) * 3600 - now.minute * 60 - now.second
+            builder.master.reactor.callLater(
+                delay,
+                builder.buildset_manager.submitBuildSet,
+                requests[0]
```
picnixz, line 302:
```suggestion
                requests[0],
```

Just for consisetncy across the file.

master/custom/factories.py
```diff
@@ -259,6 +259,12 @@ class UnixBuildWithoutDocStrings(UnixBuild):
     configureFlags = ["--with-pydebug", "--without-doc-strings"]
 
 
+class UnixBigmemBuild(UnixBuild):
+    buildersuffix = ".bigmem"
+    testFlags = ["-M60g", "-j4", "-uall,extralargefile"]
+    factory_tags = ["aarch64", "bigmem"]
```
zware, line 265: The `aarch64` tag doesn't really belong here; there's nothing specific to `aarch64` in this builder factory (it would work perfectly well with a similarly beefy x86_64 machine).  We do need to fix how tags are assigned to include tags from the worker as well, though.

master/master.cfg
```diff
@@ -315,18 +333,24 @@ for name, worker_name, buildfactory, stability, tier in BUILDERS:
...
+        locks=[cpulock.access("counting")],
     )
 
+    # This worker runs pyperformance at 12am. If a build is scheduled between
+    # 10pm and 2am, it will be delayed at 2am.
+    if worker_name == "diegorusso-aarch64-bigmem":
+        builder.nextBuild = no_builds_between(22, 2)
+
+    c["builders"].append(builder)
+
```
zware, line 353: Note that this section is specific to the PR builders; for regular builders look around line 250.  The same treatment will be needed there.

master/custom/workers.py
zware: That is an option, but has different implications.  I suspect this is following how the other bigmem worker is configured, which is set to allow the beta branch to be built without having to change anything extra when it is created.

master/custom/factories.py
diegorusso: ok, I'll remove it then.

master/custom/builders.py
diegorusso: Moved, thanks!

master/master.cfg
diegorusso: OK, replicated the same logic in the previous section.

zware:
Please note, GitHub's review interface is useless with force-pushes.  This repository does squash merges just like `cpython`, so it's better to just push new commits with review changes.

master/custom/workers.py
```diff
@@ -144,6 +144,12 @@ def get_workers(settings):
...
+            parallel_tests=60,
```
zware, line 151: Just noticing this value, it seems a little extreme for a bigmem builder.  Theoretically, this could call on 3600GB of RAM (60 tests x 60GB each from the `-M60g` option).  In practice there aren't enough bigmem tests to *actually* request that, but this is a big enough number of tests that it becomes likely that multiple (possibly *all*) bigmem tests will be running in parallel.

master/master.cfg
```diff
@@ -182,6 +182,24 @@ def is_important_change(change):
     return any(is_important_file(filename) for filename in change.files)
 
 
+# Avoid a build to be started between start and end time and delay such build
+# at end time
+def no_builds_between(start, end):
+    def f(builder, requests):
+        now = datetime.now()
+        if start <= now.hour < end:
```
zware, line 190: This condition will never be true with `start=22` and `end=2`.

diegorusso:
> Please note, GitHub's review interface is useless with force-pushes. This repository does squash merges just like `cpython`, so it's better to just push new commits with review changes.

I know, the force push were to fix a rebase I did on top of main. I really don't like the merge commit that Github does when "updating" the branch when outdated. IIRC the other force push  was to fix a typo in the commit message.
Anyway, I'll be more careful for those as I'm aware that force pushes are not greater (in general and) in Github.

master/master.cfg
diegorusso: You are right. I  wrote the prototype but then left as is. I'll fix it tomorrow.

master/custom/workers.py
diegorusso: Yes, my bad. Decreased to 4, so we should be good (60*4=240)

master/master.cfg
diegorusso: Hopefully now the code is better to read and understand. Now it works when the time range span over midnight.

zware (approved):
LGTM.

One question, though: is there any way to do the scheduling trickery at the worker level rather than the builder level?  I feel like that might be a cleaner long term solution, but as I'm not sure it's possible I'm not going to push for it :).  If we do determine later that it's doable (or we have issues with what's here) we can always change it later.

zware:
Deployed! https://buildbot.python.org/#/workers/114

diegorusso:
> LGTM.
> 
> One question, though: is there any way to do the scheduling trickery at the worker level rather than the builder level? I feel like that might be a cleaner long term solution, but as I'm not sure it's possible I'm not going to push for it :). If we do determine later that it's doable (or we have issues with what's here) we can always change it later.

Thanks for merging the PR. The worker is now online! https://buildbot.python.org/#/workers/114

Regarding your comment, I agree with you we should do this at the worker level but I'm no expert in buildbot and this is *a* way I found to do it. If anyone out there could point me to the right direction, I'll be more than happy to do it.

master/master.cfg
```diff
@@ -182,6 +182,46 @@ def is_important_change(change):
...
+    now = datetime.now().time()
+    start = datetime.strptime(start, "%H:%M").time()
+    end = datetime.strptime(end, "%H:%M").time()
+    def f(builder, requests):
+        if is_within_time_range(now, start, end):
+            delay = get_delay(now, end)
+            # Schedule the build later
+            builder.master.reactor.callLater(
+                int(delay),
+                builder.buildset_manager.submitBuildSet,
```
encukou, line 216: We're getting an exception here:
```
    Traceback (most recent call last):
      File "/srv/buildbot/venv/lib/python3.9/site-packages/twisted/internet/defer.py", line 980, in _startRunCallbacks
        self._runCallbacks()
      File "/srv/buildbot/venv/lib/python3.9/site-packages/twisted/internet/defer.py", line 1074, in _runCallbacks
        current.result = callback(  # type: ignore[misc]
      File "/srv/buildbot/venv/lib/python3.9/site-packages/twisted/internet/defer.py", line 1960, in _gotResultInlineCallbacks
        _inlineCallbacks(r, gen, status, context)
      File "/srv/buildbot/venv/lib/python3.9/site-packages/twisted/internet/defer.py", line 2014, in _inlineCallbacks
        result = context.run(gen.send, result)
    --- <exception caught here> ---
      File "/srv/buildbot/venv/lib/python3.9/site-packages/buildbot/process/buildrequestdistributor.py", line 262, in _getNextUnclaimedBuildRequest
        nextBreq = yield self.nextBuild(self.bldr, breqs)
      File "/srv/buildbot/master/master.cfg", line 216, in f
        builder.buildset_manager.submitBuildSet,
    builtins.AttributeError: 'Builder' object has no attribute 'buildset_manager'
```

encukou: If I'm reading the [nextBuild](https://docs.buildbot.net/4.1.0/manual/configuration/builders.html#:~:text=nextBuild) and Twisted docs correctly, this should return a Deferred like this:
```py
from twisted.internet import defer
...
            deferred = defer.Deferred()
            builder.master.reactor.callLater(
                int(delay),
                deferred.callback,
                requests[0],
            )
            return deferred
```
But, it would be nice to get confirmation from someone who's used Twisted before :)

encukou: I filed #553

PR #547, opened 2024-10-25
Improve the Release status dashboard

encukou:
I've spent some time making the Release Dashboard closer to my daily status reports (and include the info needed for them). Here's the result.

image

Outdated static demo: https://encukou.github.io/buildbot-status-dashboard/
WIP repo, with caching and hacks needed to iterate on this locally: https://github.com/encukou/buildbot-status-dashboard

Technically it *should* be drop-in upgrade for the existing dashboard, and that's what this PR does. But I now think it would be better to have both, at least for a while, in case this falls over. But, code is here.

Does this look reasonable?

---

Quick branch info (status, RM name, release schedule link) for all branches at the top, with detailed per-branch sections after that.
Details can generally be expanded/collapsed.

Show tier-3+ failures: these aren't considered release-blocking but we tend to fix these ASAP. These mark a "green" branch/builder as "yellow".

Show low-priority problems: disconnected builders, build warnings, unstable builder failures. These don't affect the overall status.

Show changes and test results for failing builds (when available).

Show build history overview for builders; this is useful context.

Conceptually, the branch sections list "problems", not builders, in preparation for adding blocker issues from GitHub. Problems are sorted by severity & grouped by type; the highest-severity problem determines the overall branch status.

Pull Request builders are, for now, shown in a `no-branch` section (hiding at the end with no link to it).

Based on Pablo's existing dashboard and his BuildbotHammer tool.

### DIFF
--- a/master/custom/release_dashboard.py
+++ b/master/custom/release_dashboard.py
@@ -33,7 +33,7 @@ BRANCHES_URL = "https://raw.githubusercontent.com/python/devguide/main/include/r
 
 
 def _gimme_error(func):
-    """Decorator to turn AttributeError into a different Exception
+    """Debug decorator to turn AttributeError into a different Exception
 
     jinja2 tends to swallow AttributeError or report it in some place it
     didn't happen. When that's a problem, use this decorator to get
@@ -43,8 +43,10 @@ def _gimme_error(func):
         try:
             return func(*args, **kwargs)
         except AttributeError as e:
-            raise Exception(f'your error: {e!r}')
+            raise _WrappedAttributeError(f'your error: {e!r}')
     return decorated
+
+class _WrappedAttributeError(Exception): pass
 
 
 class DashboardObject:
@@ -307,13 +309,10 @@ class Branch(_BranchTierBase):
 
     @cached_property
     def featured_problem(self):
-       try:
         try:
             return self.problems[0]
         except IndexError:
             return NoProblem()
-       except AttributeError:
-           raise SystemError
 
     def get_grouped_problems(self):
         def key(problem):
@@ -411,18 +410,35 @@ class Build(DashboardObject):
 
     @cached_property
     def junit_results(self):
-        filepath = (
-            self._root._app.test_result_dir
-            / self.builder.branch.tag
-            / self.builder["name"]
-            / f'build_{self["number"]}.xml'
-        )
-        try:
-            file = filepath.open()
-        except OSError:
+        if not self._root._app.test_result_dir:
             return None
-        with file:
-            etree = ElementTree.parse(file)
+
+        try:
+            filepath = (
+                self._root._app.test_result_dir
+                / self.builder.branch.tag
+                / self.builder["name"]
+                / f'build_{self["number"]}.xml'
+            ).resolve()
+
+            # Ensure path doesn't escape test_result_dir
+            if not filepath.is_relative_to(self._root._app.test_result_dir):
+                return None
+
+            if not filepath.is_file():
+                return None
+
+            with filepath.open() as file:
+                etree = ElementTree.parse(file)
+
+        # We don't have a logger set up, this returns None on common failures
+        # (meaning failures won't show on the dashboard).
+        # TODO: set up monitoring and log failures (in the whole method).
+        except OSError as e:
+            return None
+        except ElementTree.ParseError as e:
+            return None
+
         result = JunitResult(self, {})
         for element in etree.iterfind('.//error/..'):
             result.add(element)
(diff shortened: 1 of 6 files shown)


master/custom/release_dashboard.py
```diff
@@ -1,122 +1,730 @@
...
-        disconnected_workers = {}
+    jinja2 tends to swallow AttributeError or report it in some place it
+    didn't happen. When that's a problem, use this decorator to get
+    a usable traceback.
+    """
+    def decorated(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except AttributeError as e:
+            raise Exception(f'your error: {e!r}')
```
pablogsal, line 46: Maybe raise something more concrete (maybe `TemplateAttributeError` or something else).

master/custom/release_dashboard.py
```diff
@@ -1,122 +1,730 @@
...
-            disconnected_workers=sorted(disconnected_workers.items()),
+    @cached_property
+    def featured_problem(self):
+       try:
+        try:
+            return self.problems[0]
+        except IndexError:
+            return NoProblem()
+       except AttributeError:
+           raise SystemError
```
pablogsal, line 316: Why `SystemError`? Shouldn't better be something concrete like`BuilderError` or something akin to that?

master/custom/release_dashboard.py
```diff
@@ -1,122 +1,730 @@
...
+        try:
+            file = filepath.open()
+        except OSError:
+            return None
+        with file:
+            etree = ElementTree.parse(file)
+        result = JunitResult(self, {})
+        for element in etree.iterfind('.//error/..'):
+            result.add(element)
         return result
```
pablogsal, line 429: (This assumes there is a logger somewhere, but you can drop that)

```suggestion
@cached_property
def junit_results(self):
    if not self._root._app.test_result_dir:
        return None
        
    try:
        filepath = (
            self._root._app.test_result_dir
            / self.builder.branch.tag
            / self.builder["name"]
            / f'build_{self["number"]}.xml'
        ).resolve()
        
        # Ensure path doesn't escape test_result_dir
        if not str(filepath).startswith(str(self._root._app.test_result_dir)):
            logger.warning(f"Attempted path traversal: {filepath}")
            return None
            
        if not filepath.is_file():
            return None
            
        with filepath.open() as file:
            etree = ElementTree.parse(file)
            
    except OSError as e:
        logger.error(f"Failed to read JUnit results: {e}")
        return None
    except ElementTree.ParseError as e:
        logger.error(f"Failed to parse JUnit XML: {e}")
        raise XMLParseError(f"Malformed JUnit XML: {e}") from e
```

master/custom/release_dashboard.py
```diff
@@ -1,122 +1,730 @@
...
+            name_parts.pop(0)
+        for part in name_parts:
+            result.error_types.update(error_types)
+            result = result.contents.setdefault(part, JunitResult(self, {}))
+        result.error_types.update(error_types)
+        for error in errors:
+            if error not in result.errors:
+                # De-duplicate, since failing tests are re-run and often fail
+                # the same way
+                result.errors.extend(errors)
```
pablogsal, line 473: Hummmm aren'y you adding technically all the errors again and again once per error to the list? What am I missing?

pablogsal (commented):
Some comments for a first pass

pablogsal:
Wow Petr, this is absolutely fantastic work! 🎉

I'm really excited to see the release dashboard coming together like this! This is going to make everyone's lives so much easier.

I made a first pass but given how big this PR is I probably missed stuff so I may do one more later.

master/custom/release_dashboard.py
encukou: Sure.
(This is a debug helper -- but one that I find myself using constantly so I'd rather leave it in.)

encukou (commented):
Thank you for your kind words!

Do you think it's OK to replace the existing dashbuard, or should we leave them side-by-side for a while?

vstinner:
> Do you think it's OK to replace the existing dashboard, or should we leave them side-by-side for a while?

If it's quick to load, IMO it's ok to replace the existing one.

encukou:
I've been using this for a week, and saw a minor display issue (and many out-of-sync results as the local version needs heavy caching).
Integrating with the live buildbot might be a bit rough, of course.

I plan to merge this around Wednesday (if there are no objections) and iterate from there.

> If it's quick to load, IMO it's ok to replace the existing one.

It should be on the same order of magnitude as the current dashboard.
Quite slow, but cached, and re-generated by a cron job every 5 minutes.

pablogsal:
I'm currently on vacation for two weeks so unfortunately I was not able to find the time to review this again so feel free to land it if you are happy with it and we can fix problems later :)

vstinner:
Thanks for this great enhancement Petr!

encukou:
Thanks for the help & reviews!
It's up now and works, knock on wood :)

PR #553, opened 2024-11-06
Fixes for diegorusso-aarch64-bigmem

encukou:
Follow-up to #546

- Get the current time when a build is requested, rather than when the config is first loaded
- Return a Deferred from nextBuild. By my reading of the [`nextBuild`](https://docs.buildbot.net/4.1.0/manual/configuration/builders.html#:~:text=nextBuild) and [`Deferred`](https://docs.twisted.org/en/stable/core/howto/defer.html) docs, this is the way to do it.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -20,6 +20,7 @@ from functools import partial
 from buildbot.plugins import reporters, schedulers, util
 from buildbot import locks
 from twisted.python import log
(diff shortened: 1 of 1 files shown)


encukou:
`checkconfig` & `reconfig` [works](https://buildbot.python.org/all/#/builders/1618/builds/11).

PR #554, opened 2024-11-06
Increase timeout for slow tests on UnixBigMemBuild

diegorusso:
`test_lzma` and `test_bigmem` require more time to be fully executed. Increase the default timeout to allow these tests to pass.


diegorusso:
On my worker:
* test_lzma passed: 33 min 39 sec
* test_bigmem: 1 hour 6 min

vstinner:
Merged, thanks.

PR #556, opened 2024-11-07
Fix WASI buildbots to support the new default target

brettcannon:
Also make the target more explicit.


brettcannon:
Waiting on https://github.com/python/cpython/pull/126561 to land before taking this out of draft.

PR #557, opened 2024-11-15
Promote diegorusso-aarch64-bigmem to stable Tier-2

diegorusso:
The worker has been stable for almost a couple of weeks with continous daily builds.


zware (approved):

diegorusso:
if we want to wait exact 14 days, we can merged it next week :)

zware:
No need :)

PR #558, opened 2024-11-18
FIPS builders should compile only 3.12+ branches

stratakis:


stratakis:
@encukou

stratakis:
3.12+ branches work better with FIPS since https://github.com/python/cpython/pull/20380/files

As the multiprocessing fix is quite invasive for backport, it makes sense to keep testing FIPS on 3.12+

vstinner (approved):
LGTM.

@encukou: Would you mind to double check?

encukou:
Thank you, this looks like the best solution!

Today is release day -- a bad time to restart Buildbot. I'll test & merge after the release.

PR #561, opened 2024-11-22
Add Windows PGO builder

zware:
- **Add Windows64PGOBuild factory**
- **Add Windows PGO builder to main branch**
- **Remove a couple of dead code branches**
- **Allow multiple workers per builder**
- **Add ware-win11 to Windows PGO builder**

### DIFF
--- a/master/custom/builders.py
+++ b/master/custom/builders.py
@@ -328,7 +328,7 @@ def get_builders(settings):
 
 
 # Match builder name (excluding the branch name) of builders that should only
-# run on the main and "custom" branches.
+# run on the main and PR branches.
 ONLY_MAIN_BRANCH = (
     "Alpine Linux",
     "ARM64 Windows",
--- a/master/custom/builders.py
+++ b/master/custom/builders.py
@@ -35,6 +35,7 @@ from custom.factories import (
     Windows64Build,
     Windows64BigmemBuild,
     Windows64NoGilBuild,
+    Windows64PGOBuild,
     Windows64RefleakBuild,
     Windows64ReleaseBuild,
     MacOSArmWithBrewBuild,
@@ -219,6 +220,9 @@ UNSTABLE_BUILDERS_TIER_1 = [
     ("AMD64 CentOS9 FIPS No Builtin Hashes", "cstratak-CentOS9-fips-x86_64", CentOS9NoBuiltinHashesUnixBuild),
 
     ("AMD64 Arch Linux Valgrind", "pablogsal-arch-x86_64", ValgrindBuild),
+
+    # Windows MSVC
+    ("AMD64 Windows PGO", "bolen-windows10", Windows64PGOBuild),
 ]
 
 
@@ -330,6 +334,7 @@ ONLY_MAIN_BRANCH = (
     # Cygwin is not supported on 2.7, 3.6, 3.7
     "Cygwin",
     "ARM64 Windows",
+    "Windows PGO",
     "AMD64 Arch Linux Perf",
     "AMD64 Arch Linux Valgrind",
 )
--- a/master/custom/builders.py
+++ b/master/custom/builders.py
@@ -331,8 +331,6 @@ def get_builders(settings):
 # run on the main and "custom" branches.
 ONLY_MAIN_BRANCH = (
     "Alpine Linux",
-    # Cygwin is not supported on 2.7, 3.6, 3.7
-    "Cygwin",
     "ARM64 Windows",
     "Windows PGO",
     "AMD64 Arch Linux Perf",
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -666,6 +666,13 @@ class Windows64ReleaseBuild(Windows64Build):
     factory_tags = ["win64", "nondebug"]
 
 
+class Windows64PGOBuild(Windows64ReleaseBuild):
+    buildersuffix = ".pgo"
+    buildFlags = Windows64Build.buildFlags + ["--pgo"]
+    testFlags = [*Windows64Build.testFlags, "+d"]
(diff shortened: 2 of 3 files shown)


db3l:
Heh, so this beat me by a few minutes of my own attempt.  Still trying to figure out this review process, but based on some quick local testing, I believe I'm only seeing PGO builders for me and not you in the generated config with these changes.  But I need to make sure I'm not doing something dumb trying to examine the config locally.

Oh, and also in both our cases, a PGO builder is being created for the pull_request branch, which I don't think we need, but I haven't looked yet to see why that's not filtered out by the ONLY_MAIN_BRANCH stuff.

db3l:
Ah, ok, so the issue is my quick checked filtered by workerbuilddir, and this change ends up assigning the first worker build dir to all the workers in the list.  So while the ware-win11 worker was included, it arguably had the wrong local build dir (which would be bolen-windows10).  Perhaps that's why the existing code stuck to a single worker per BuilderConfig call.  I suppose it's debatable as to why worker build dirs are based on the worker name, but that's obviously got a long history.

I'd probably just stay with the current structure and use an entry per worker for the PGO builders.  Unless you wanted to modify the change to iterate over worker_names keeping individual BuilderConfig calls per worker (and worker dir).

Oh, and clearly the pull_request branch is just handled later, and doesn't have the ONLY_MAIN_BRANCH check the main builder loop does.  Not sure why as I'd think something in ONLY_MAIN_BRANCH should be, well, only the main branch.  But adding that check to the pull_request loop as well would clearly be backwards-incompatible, though I wonder if anyone wants pull_requests to run on those non-main-branch builders?

zware:
What I'm going for here is to actually share the workload of the builder between the two workers as a proof of concept for a long-term goal of mine to simplify things to a tag-based system where we can just set things up with the right tags and the configuration will match up builders with workers for us.  That and I don't think we need two new PGO builders when two workers can share the load :)

I agree that the use of `worker_name` for other bits within the branch is less than ideal, and I specifically put yours first in the list to make it the "primary" here because I can deal with any fallout on my worker directly.

The builddir name matching the worker name is a historical artifact, and doesn't really mean much.  If we had used the builder name (or something based on it) in the first place, we wouldn't have had to add the 'buildersuffix' hack later on when we added extra builders to workers :).  With that in mind, what do you think of this latest commit, which will hopefully work as expected?

db3l:
Oh, I definitely side with you on the goals.  I'm just not sure I'd mix those changes with the builder addition for the PGO issue, at least not in a backwards incompatible way.  I mean, the definitions change infrequently, but someone might take advantage in a future change, only to find something not working that used to (especially if converting an existing builder entry), such as a worker getting wrong or missing extra factory settings.

For build directory naming, I'd probably separate that as well, if only due to changes on existing workers involving so much disk space, so some coordination may be needed.  In particular switching names based on a builder entry being singleton vs tuple can result in unexpected disk usage on an existing worker following an update.

With that said, here's what I get for builders with the current code:

```
>>> pprint.pprint([(x.name,x.workernames,x.workerbuilddir,x.builddir) 
                   for x in b if 'Windows PGO' in x.name])
[('AMD64 Windows PGO 3.x',
  ['bolen-windows10', 'ware-win11'],
  '3.x.bolen-windows10.pgo',
  '3.x.bolen-windows10.pgo'),
 ('AMD64 Windows PGO PR',
  ['bolen-windows10', 'ware-win11'],
  'PullRequest-nondebug-pgo-tier-1-unstable-win64',
  'PullRequest-nondebug-pgo-tier-1-unstable-win64')]
````
The main branch builddir is unchanged (I think you have a dangling lambda there - you're missing the "builddir=").

The PGO dir is annoyingly ugly and long, but I suppose any new naming scheme will have bike-shedding.  But perhaps introduce the naming change consistently at some point rather than just here.

I'm still wondering if main branch only builders should also be excluded from pull_request builds in general, though that would be a behavior change.  But I'm not sure any of those excluded builders would anticipate being used for pull requests too.

Perhaps consider just a nested loop in master.cfg rather than fiddling with worker_name/worker_names throughout the body?  That is, for the two loops (branches and pull requests), something like:

```
for name, worker_names, buildfactory, stability, tier in BUILDERS:
    for worker_name in (worker_names if isinstance(worker_names,tuple) else [worker_names]):
        ...existing body...
```

That way, everything should (I believe) work just as today (parallel/extra parameters, builder configuration values) regardless of builder list entry, and existing entries can be converted to use a list safely with no impact on existing workers.  Doing something different with build folder names and removing the need for extra suffix information could be a subsequent step.

zware:
> The main branch builddir is unchanged (I think you have a dangling lambda there - you're missing the "builddir=").

:man_facepalming: 

> That way, everything should (I believe) work just as today (parallel/extra parameters, builder configuration values) regardless of builder list entry, and existing entries can be converted to use a list safely with no impact on existing workers. Doing something different with build folder names and removing the need for extra suffix information could be a subsequent step.

That can't actually work; the "extras" are actually settings for the builder rather than for the worker, even though they're stored per-worker.  We've historically not separated the two very well, with the assumption that any builder would only ever be run on one worker.

I'll just revert the attempt at a shared builder for now and go ahead with just adding the builder to your worker.

db3l:
> That can't actually work; the "extras" are actually settings for the builder rather than for the worker, even though they're stored per-worker. We've historically not separated the two very well, with the assumption that any builder would only ever be run on one worker.

Ah, yeah, I see, so even with an inner loop you still couldn't use the new notation anyway with workers with incompatible parallel or extra settings, which might not be obvious.  Phooey, foot gun.

Bummer.  And sorry for the negativity.  I'd be happy to help contribute to a subsequent process to try to clean some of this up though.

Issue #560, opened 2024-11-22
Add PGO builds on Windows to buildbots


mdboom:
# Feature or enhancement

### Proposal:

We currently don't have any build bot coverage that tests PGO on Windows.  This means we land code that breaks the Windows release build not all that infrequently (examples over the last 12 months: python/cpython#113655, python/cpython#125217, python/cpython#123153, python/cpython#122921).

It would be nice to add this to existing buildbots, but I know our Windows buildbots are few, and PGO builds are time consuming.  Cc'ing the Windows buildbot owners for discussion: @ambv, @itamaro, @db3l.

As an aside, there is plenty of [PGO coverage on the Linux buildbots](https://buildbot.python.org/#/builders?tags=%2Bpgo), but that's largely irrelevant to the problem as it's an entirely different compiler toolchain.  There is no PGO coverage on macOS either, but I have never seen that break -- if we want that for completeness, I suggest we file another issue to track that.

### Has this already been discussed elsewhere?

No response given

zware:
I've opened #561 for this.  I'm trying something new in sharing this builder between both of our workers, @db3l.  It's a hack, but hopefully it works out :)

zware:
Done: https://buildbot.python.org/#/builders/1620

mdboom:
Thanks, all!

itamaro:
opened #565 to add a Windows PGO NoGIL builder

PR #565, opened 2024-11-24
gh-560: Add Windows PGO NoGIL builder

itamaro:


itamaro:
@zware @mdboom is there are no concerns with the implementation, I will merge this tomorrow, when I can keep an eye on the Windows worker more closely.

db3l:
@itamaro, if you could, when you enable this tomorrow, could you specifically take a peek at how test_cext does?

It looks like it's been failing on my new PGO builder.  I haven't had a lot of time to look into it, but from what I can tell it didn't run in the past as it skips debug mode on Windows - but the PGO builder is no longer debug mode.  I didn't notice this during my manual test as I hadn't enabled the "cpu" resource so it didn't even try test_cext.

What I'm not sure is whether test_cext should be modified to skip any Windows build, or just PGO, or if it ought to work in release/pgo mode and the PGO builder is just highlighting something on my worker.

vstinner (approved):
LGTM

mdboom (approved):
LGTM

itamaro:
@db3l `test_cext` passed, but 5 other tests failed (test_call test_class test_functools test_int test_ssl)

logs: https://buildbot.python.org/#/builders/1622/builds/1/steps/5/logs/stdio

PR #569, opened 2025-01-22
Revert "run make regen-requirements (#568)"

hugovk:
This reverts commit 49673fe41b5e1e551e37a3ad83fe9cdab81b189d from https://github.com/python/buildmaster-config/pull/568.

The `test-with-buildbots` label hasn't bee triggering the fleet, for example for https://github.com/python/cpython/pull/129140

`twistd.log` has `builtins.DeprecationWarning: buildbot.util.httpclientservice.HTTPClientService.post was deprecated in buildbot 4.1.0; please use Use HTTPSession.post() instead`

So let's try a revert to see if this fixes it.


pablogsal:
See https://github.com/python/buildmaster-config/pull/574

PR #571, opened 2025-01-22
Link to buildbot results in scheduling message

hugovk:
Fixes https://github.com/python/buildmaster-config/issues/327.

Issue #327, opened 2022-06-07
Add link to buildbot results


pitrou:
When the "test-with-buildbots" label is applied, Bedevere [posts a comment](https://github.com/python/cpython/pull/30617#issuecomment-1148319780) to notify that builds have been scheduled, but the comment is quite unhelpful as it doesn't say how the results can be accessed.

It should be possible to add a hyperlink to the buildbot results page for the particular commit, or at least the given PR (for example https://buildbot.python.org/all/#/grid?branch=refs%2Fpull%2F30617%2Fmerge or https://buildbot.python.org/all/#/changes/14341 in the case above).

pablogsal:
> When the "test-with-buildbots" label is applied, Bedevere [posts a comment](https://github.com/python/cpython/pull/30617#issuecomment-1148319780) to notify that builds have been scheduled, but the comment is quite unhelpful as it doesn't say how the results can be accessed.

But I think adding a link to https://buildbot.python.org/all/#/grid?branch=refs%2Fpull%2F30617%2Fmerge is a good idea that we can explore 👍

hugovk:
Please see PR https://github.com/python/buildmaster-config/pull/571.

PR #576, opened 2025-01-24
Link to PR-specific grid in PR message

zware:
Also add the link to 'command' PR runs rather than just 'label' runs.

Expansion of #571; ref #327.

### DIFF
--- a/master/custom/pr_testing.py
+++ b/master/custom/pr_testing.py
@@ -15,19 +15,20 @@ REFLEAK_TESTING_LABEL = ":hammer: test-with-refleak-buildbots"
 
 GITHUB_PROPERTIES_WHITELIST = ["*.labels"]
 
-BUILD_SCHEDULED_MESSAGE_TEMPLATE = """\
+BUILD_MESSAGE_HEADER = """\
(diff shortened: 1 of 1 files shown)


vstinner (approved):
It seems to be good to me, but I didn't try the URL :-)

hugovk:
Ah right, I see my previous https://github.com/python/buildmaster-config/pull/571 linked to an old PR number, thanks for this update.

Let's check the link.

Here's a PR that used buildbots:

https://github.com/python/cpython/pull/129252

Here's a URL with that PR number:

https://buildbot.python.org/all/#/grid?branch=refs%2Fpull%2F129252%2Fmerge

It shows this, but I don't see "gh-129248: Filter out the iOS log prefix from testbed runner output"?

image

What's the expected result?

zware:
Finally getting back to this, I'm afraid we can't actually link to the particular branch in the grid directly (anymore? I've opened buildbot/buildbot#8324).  Maybe we could use https://buildbot.python.org/all/#/waterfall?tags=PullRequest, but that's a very different view than the grid, and less specific as well.

hugovk (approved):
Well, this is still an improvement, thanks!

PR #577, opened 2025-02-16
Add cache configuration

zware:
These are mostly made up numbers just to see if there is any noticeable effect on either performance or memory usage.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -110,6 +110,11 @@ c["configurators"] = [
     )
 ]
 
+# Note: these cache values are not currently tuned in any meaningful way.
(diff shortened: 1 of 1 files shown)


master/master.cfg
```diff
@@ -110,6 +110,17 @@ c["configurators"] = [
     )
 ]
 
+c["caches"] = {
```
vstinner, line 113: Can you please add a comment explaining how you chose these numbers?

I'm fine with "These are mostly made up numbers just to see if there is any noticeable effect on either performance or memory usage."

vstinner (approved):
LGTM

PR #584, opened 2025-03-18
Use canStartBuild rather than nextBuild for avoiding perf runs

encukou:
There is some evidence that a delay in nextBuild can leak to unrelated workers.

This switches to canStartBuild, whose docs show an example to put the worker in “quarantine” and avoid the quarantine's exponential backoff mechanism.

I'd welcome some eyes on this before testing on the live box.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -223,21 +223,22 @@ def get_delay(now, end):
 def no_builds_between(start, end):
     start = datetime.strptime(start, "%H:%M").time()
     end = datetime.strptime(end, "%H:%M").time()
-    def f(builder, requests):
+    @defer.inlineCallbacks
+    def canStartBuild(builder, wfb, request):
         now = datetime.now().time()
         if is_within_time_range(now, start, end):
             delay = get_delay(now, end)
-            # Schedule the build later
-            deferred = defer.Deferred()
-            builder.master.reactor.callLater(
-                int(delay),
-                deferred.callback,
-                requests[0],
-            )
-            return deferred
+            # Adapted from: https://docs.buildbot.net/current/manual/customization.html#canstartbuild-functions
+            wfb.worker.quarantine_timeout = delay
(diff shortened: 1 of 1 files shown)


master/master.cfg
```diff
@@ -223,21 +223,22 @@ def get_delay(now, end):
...
+    def canStartBuild(builder, wfb, request):
         now = datetime.now().time()
         if is_within_time_range(now, start, end):
             delay = get_delay(now, end)
-            # Schedule the build later
-            deferred = defer.Deferred()
-            builder.master.reactor.callLater(
-                int(delay),
-                deferred.callback,
-                requests[0],
```
diegorusso, line 235: I think the problem is here, where the first element of requests is a "random" job that gets deferred.

diegorusso (commented):
Thanks for the change, it looks OK to me.

diegorusso (approved):

zware (approved):
This looks fine to me.  I do wonder if `pause` rather than `quarantine` might be a cleaner solution overall (ideally controlled by the worker itself; I have a question in to the `#buildbot` IRC channel about whether there's a nice way to do that), but I think this is improvement enough to go ahead with it anyway.

master/master.cfg
```diff
@@ -223,21 +223,22 @@ def get_delay(now, end):
 def no_builds_between(start, end):
     start = datetime.strptime(start, "%H:%M").time()
     end = datetime.strptime(end, "%H:%M").time()
-    def f(builder, requests):
+    @defer.inlineCallbacks
```
encukou, line 226:
```suggestion
```
```
	builtins.TypeError: inlineCallbacks requires <function no_builds_between.<locals>.canStartBuild at 0x7847de000c10> to produce a generator; instead got False
```

encukou:
AFAICS, quarantine has a built-in timeout, but to un-pause a worker we'd need to reliably schedule running some un-pausing code.

PR #585, opened 2025-03-24
Add s390x Fedora Stable and Rawhide workers/builders

stratakis:


stratakis:
Already have a password for the workers

vstinner (approved):
LGTM. Let's see how it goes :-)

vstinner:
s390x Fedora Stable and s390x Fedora Rawhide 3.x latest builds are green!

PR #588, opened 2025-04-14
Add colour to Buildbot output

hugovk:
Like https://github.com/python/cpython/pull/129196, will help us find errors in logs more easily.

### DIFF
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -63,7 +63,7 @@ class UnixBuild(BaseBuild):
     interpreterFlags = ""
     testFlags = ["-j2"]
     makeTarget = "all"
-    test_environ = {}
+    test_environ = {"FORCE_COLOR": 1}
     build_out_of_tree = False
 
     def setup(self, parallel, branch, test_with_PTY=False, **kwargs):


vstinner:
I'm not convinced by this change. It makes my typical use case more painful: download logs and open them in a text editor (vim). There are now ANSI codes which make the logs harder to read :-(

> Set to draft as a demo build shows some colour output but also ANSI codes leaking out:

At the beginning, there are nice colors. At the middle, raw ANSI codes are displayed which makes the output harder to read :-(

Screenshot 2025-04-15 at 10-56-26 Buildbot

Also, if I download the logs and open them my text editor (vim), I also see the raw ANSI codes (not colors).

hugovk:
Yeah, it's not ready for merge, it's not meant to switch from colour to ANSI codes mid-way.

> Also, if I download the logs and open them my text editor (vim), I also see the raw ANSI codes (not colors).

Hmm, even if we fix the switch, the codes would still be there when downloading logs viewing in a text editor.

This is also the case with GitHub Actions: compare the [rendered logs](https://github.com/python/cpython/actions/runs/14458218939/job/40545806733) and the [raw logs](https://productionresultssa4.blob.core.windows.net/actions-results/a4fc5595-b47d-422d-a613-f0cb40a36d6a/workflow-job-run-352a2ad8-20cf-5ada-768f-832deda5209a/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-04-15T10%3A06%3A15Z&sig=kT9Mx2%2FQaxxbfQACDGS2Jhh3CGDSyJkRKKmblVpwSIk%3D&ske=2025-04-15T21%3A09%3A00Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-04-15T09%3A09%3A00Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-01-05&sp=r&spr=https&sr=b&st=2025-04-15T09%3A56%3A10Z&sv=2025-01-05). But I think people rarely look at the raw logs. (I occasionally diff them, but the rendered colour is much more beneficial to me than the codes when diffing.)

If it's more important to keep raw logs more readable than rendered logs for Buildbots, then we can just close this.

---

btw, is there a way to download the full logs from a Buildbot run (that is, all of "configure", "compile", "test", and so on), rather than just for a single step (for example, just "test")?

vstinner:
> If it's more important to keep raw logs more readable than rendered logs for Buildbots, then we can just close this.

The problem of buildbots is that the web site is very slow and buggy. Reading logs with an external tool is a workaround for that. It's also easier for me to navigate large logs (2 000 lines or more) with a fast text editor, rather than the slow web site.

> btw, is there a way to download the full logs from a Buildbot run (that is, all of "configure", "compile", "test", and so on), rather than just for a single step (for example, just "test")?

I'm not aware of such feature, I always download logs of a single step such as "test".

hugovk:
Closing as the buildbot UI is slow for rendered logs, and we often need to view the plain text logs instead.

PR #589, opened 2025-04-18
Prioritize bigmem tests on bigmem workers

ambv:
This ensures those tests run first, shortening the entire run.

The test set to prioritize is a bit different between Unix and Windows, this is based on actual observed runtimes of those tests. I took tests that take >5 minutes to execute.

Unrelatedly, mark ambv-bb-win11 as unavailable between 7:20am UTC - 9:20am UTC so that it can restart in peace without interrupting running builds.


encukou:
The maintenance window is not applied to builds requested on PRs.
I assume that wasn't intentional. I sent #701 to fix this and move this to worker configuration.

PR #590, opened 2025-04-30
Increase UNIX ASAN timeout from 20 to 35 minutes

vstinner:


vstinner:
20 minutes is not enough, example: https://buildbot.python.org/#/builders/585/builds/6836

test_math, test_statistics and test_subprocess failed with a timeout of 20 minutes.

PR #595, opened 2025-05-15
Stop testing 3.12 on `stan-raspbian`

StanFromIreland:
Tests fail due to the system encoding, latin-1. These tests have been fixed for bugfix branches (main, 3.14, 3.13), but unfortunately there is nothing to be done for 3.12 now. Therefore, it is best to remove it entirely since the fail result is guaranteed.

Request; @zware


vstinner:
Merged, thanks.

PR #596, opened 2025-05-18
Promote `stan-raspbian` to stable

StanFromIreland:
The [buildbot logs](https://buildbot.python.org/#/workers/117) are [will be in theory] green, all issues have been resolved.


StanFromIreland:
There was an interesting performance change with `SlowNonDebugUnix`, building now takes ~2x the original time, where as tests take ~.33x of the original time. Better overall, but still quite odd.

zware:
Ah, I missed that you changed the builder type as well as stability :(

StanFromIreland:
Oh sorry, I noticed that's what gps-Raspbian used so I thought it would be useful for mine too, as they were taking almost one hour. It's cut the time in half, though it is no longer a debug build, which is surprisingly faster to compile.

PR #597, opened 2025-05-18
Retire skumaran-ubuntu-x86_64 worker

zware:
Last connected 6 months ago.

Thanks to Senthil for running it while it lasted!


StanFromIreland:
@orsenthil sent a message to the buildbot mailing list, it seems he would like to restart it, reverting this.

PR #600, opened 2025-05-31
Mark the Windows NoGIL PGO builder as stable

itamaro:
It started out as unstable and was broken for a while, but it is healthy now, so let's move it to stable


itamaro:
if there are no objections, I will merge it in a day or two

PR #604, opened 2025-06-04
Allow setting custom buildfactory with local worker

emmatyping:
Currently, the local worker unconditionally uses a UnixBuild. However, when working on changes to other factories, it is useful to be able to customize this in settings.yaml to run the builder locally. This PR adds a setting to allow modifying the factory used, while falling back to UnixBuild if nothing is set.

This was useful while working on https://github.com/python/buildmaster-config/pull/601


vstinner (approved):
LGTM

zware (approved):

PR #606, opened 2025-06-04
Take two at fixing WASI config.site usage

emmatyping:
This is both closer to what was done originally, and should be much more robust :)

My local test is still running, will move to ready when it is done.

The `self.compile_environ` is used during the configure here: https://github.com/python/buildmaster-config/blob/676f2e3c89117cf75dcd61ccd8c1049afb9b5512/master/custom/factories.py#L797-L804

### DIFF
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -893,6 +893,7 @@ class Wasm32WasiCrossBuild(UnixCrossBuild):
                 warnOnFailure=True,
             )
         )
+        self.compile_environ = super().compile_environ.copy()
         self.compile_environ.update(
             CONFIG_SITE=util.Interpolate("%(prop:config_site)s")
         )
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -893,7 +893,9 @@ class Wasm32WasiCrossBuild(UnixCrossBuild):
                 warnOnFailure=True,
             )
         )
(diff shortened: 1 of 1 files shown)


master/custom/factories.py
```diff
@@ -893,7 +893,9 @@ def setup(self, parallel, branch, test_with_PTY=False, **kwargs):
                 warnOnFailure=True,
             )
         )
-        self.host_configure_cmd.append(util.Interpolate("CONFIG_SITE=%(prop:config_site)s"))
+        self.compile_environ.update(
```
zware, line 896: I think we should still be defensive about mutating the class member.  I think this should do it (untested):
```suggestion
        self.compile_environ = super().compile_environ.copy()
        self.compile_environ.update(
```

zware: Possibly just `self.compile_environ.copy()` rather than `super()`.

master/custom/factories.py
```diff
@@ -893,7 +893,9 @@ def setup(self, parallel, branch, test_with_PTY=False, **kwargs):
...
+            CONFIG_SITE=util.Interpolate("%(prop:config_site)s")
```
zware, line 897: Also need to add the path prefix:
```suggestion
            CONFIG_SITE=util.Interpolate("../../%(prop:config_site)s")
```

zware (approved):
Let's try it :)

Issue #301, opened 2022-01-20
Give an example test file for /etc/buildbot/settings.yaml


emmatyping:
Hi! I am planning on adding support for cross compiling and then Emscripten as a platform. However I am unsure how to set up a realistic testing environment.

I can run the current master code, but it seems that the buildbot master relies on a file `/etc/buildbot/settings.yaml` to configure workers.

I am unsure of what the format of the file should be for testing. Could you give an example of what this file should look like please? (I can guess some of the information from settings.py, but I don't know what the configuration looks like for workers)

pablogsal:
Hi @ethanhs,

I can try to give you a sample later today but in theory, you don't need it for what you want. You may be able to just configure a local worker:

https://docs.buildbot.net/latest/manual/configuration/workers.html#local-workers

You can modify locally this:

https://github.com/python/buildmaster-config/blob/15552ea4890e4404dab4732d878dc8872ea1572a/master/master.cfg#L92

To point only to your local worker. It has been a while since I did that in the past so I am sure they may be some gotchas but that's how I have iterated in the past when doing modifications to the worker workflow or plugins.

zware:
First off, we should probably adjust https://github.com/python/buildmaster-config/blob/15552ea4890e4404dab4732d878dc8872ea1572a/master/master.cfg#L82 to check a `PYBUILDBOT_SETTINGS_PATH` environment variable for a replacement path so you don't have to stick a local test config file in `/etc/buildbot`.  From there, your test settings file could just be `use_local_worker: true` to use the single local worker, and you can add anything else you need as you find you need it :).  The only worker configuration you might need would be something like:

```yaml
workers:
    smith-emscripten:
        password: somepassword
```

ethanhs:
Ok, thank you! I will try that out. I do think having a configurable path with an example would be good, because it would prevent mistakes like committing temporary changes to `master.cfg`.

PR #607, opened 2025-06-09
Add a short guide for testing changes locally

emmatyping:
When working on the WASI builder I had to dig through the Buildbot docs and figure some things out about how to run Buildbot locally. I also wanted to document the process for using the customization of the local build factory I added in https://github.com/python/buildmaster-config/pull/604.

Fixes https://github.com/python/buildmaster-config/issues/301
(it only took me 3 years to write this up 😅  )

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ buildbot/
 lib/
 venv/
 worker/
+settings.yaml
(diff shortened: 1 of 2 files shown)


zware (approved):
LGTM, thanks :)

PR #608, opened 2025-06-12
Upload results XML for warning stable builds

encukou:
Would it make sense to save the JUnit result file for warnings builds, at least for stable builders on the main branch?

This would allow the dashboard to list tests that failed first, then succeeded on the re-run. If that happens in a test that isn't traditionally flaky, it would be good to investigate it.

---

As for this PR, I don't think this is the proper way to do things, but, I'm not sure what would be a good way. Any pointers would be welcome.

### DIFF
--- a/master/custom/steps.py
+++ b/master/custom/steps.py
@@ -116,9 +116,16 @@ class Test(BaseTest):
     }
 
     def evaluateCommand(self, cmd):
+        result = super().evaluateCommand(cmd)
         if cmd.didFail():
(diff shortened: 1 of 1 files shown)


vstinner (commented):
Which service is going to consume (remove) these XML files? In the past, these XML files filled the disk.

vstinner:
> There's a daily cronjob that removes anything from /data/www/buildbot/test-results that is a file older than 1 week

Oh ok, I'm fine with that.

PR #609, opened 2025-06-13
Fix the dashboard

encukou:
The code for finding failing streaks has a bug where it can ignore a failure. Fix that.

Also, show “green” builders in the report, so that if a similar issue comes up in the future, it can be found more easily.


encukou:
Tested on the live system; 3.9-3.11 are indeed red.
I'll merge on Monday if there are no objections.

zware (approved):

PR #614, opened 2025-07-20
Emscripten: Set PYTHON_NODE_VERSION environment variable to 24

hoodmane:
See:
https://github.com/python/cpython/issues/136852


ambv:
Let's wait until the EuroPython sprint is over before merging this as this change will cancel all running buildbot jobs.

PR #615, opened 2025-07-22
Add Emscripten libmpdec build step

hoodmane:
Goes together with:
https://github.com/python/cpython/pull/136997

### DIFF
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -1365,8 +1365,13 @@ class EmscriptenBuild(BaseBuild):
                 env=compile_environ,
             ),
             Compile(
-                name="Compile host libFFI",
+                name="Compile host libffi",
                 command=["python3", "Tools/wasm/emscripten", "make-libffi"],
+                env=compile_environ,
+            ),
+            Compile(
+                name="Compile host libmpdecimal",
+                command=["python3", "Tools/wasm/emscripten", "make-mpdec"],
                 env=compile_environ,
             ),
             Configure(


zware:
Can these not be made already present on the host?

hoodmane:
Building libmpdec is pretty fast, and it's nice for it to be reproducible. OTOH libssl takes several minutes to build...

freakboy3742:
> > Do you mean that we should build them once and then install in some prefix directory which we reuse between jobs?
> 
> Right

FWIW - this is the approach that iOS and Android use as well - we've got standalone repos ([iOS](https://github.com/beeware/cpython-apple-source-deps), [Android](https://github.com/beeware/cpython-android-source-deps)) to manage binary releases; the CPython CI builds then download/cache those releases. 

The complication for Emscripten is that we can't just add "bzip-1.0.8" and re-use it - we need to add "bzip-1.0.8, compiled for Emscripten 4.0.10". That's certainly possible, but it's definitely more complicated than the iOS/Android/macOS situation where the only real restriction is the minimum OS compatibility version.

@hoodmane One thought - we could cache these builds as part of the whole "Emscripten version cache" thing we discussed at EuroPython. 

For the benefit of others - Emscripten doesn't provide any ABI compatibility guarantees between versions. We're effectively going to have to say "Python 3.14 uses Emscripten 4.0.11", and *everything* that is for Python3.14 will need to use that version of Emscripten. 

This means the buildbot will need to maintain multiple versions of Emscripten; the idea @hoodmane and I discussed at EuroPython was to add an `--emscripten-dir` argument to the build script, and set the required Emscripten version number in the script; if `-emscripten-dir` is provided, that is used to find the appropriate EMSDK version on disk.

We could then expand that to also include cached builds of libffi, mpdecimal, or whatever else is a dependency. We'd end up with a directory structure that looks something like:
```
- emscripten-versions
  - 4.0.11
     - emsdk
     - prefix
       - include
       - lib
   - 4.0.20
     - emsdk
     - prefix
       - include
       - lib
```
The idea would be that if you specify an `--emscripten-versions` directory, it treats the content as a cache; the `build` target would assume cached builds exist and use the version specific prefix directory; targets for the build dependencies would install into the version-specific prefix.

Thoughts?

hoodmane:
Makes sense to me. libmpdec only takes about 8 seconds to build though, so could we merge it first like this and then do the rearrangement with two libraries? I feel better about having two libs before making this change. Though I'm also okay with adding `--emscripten-versions` first.

freakboy3742:
> Makes sense to me. libmpdec only takes about 8 seconds to build though, so could we merge it first like this and then do the rearrangement with two libraries? I feel better about having two libs before making this change. 

I'd be fine with merging this as is, and then re-arranging as part of the larger refactor of the build script.

StanFromIreland:
Maybe there is a way to skip the step if it is not the `3.X` or `PR` builder? Similar things have been done before I think: https://github.com/python/buildmaster-config/blob/847d2edfc6913945c21279cc0d709b86f1e6f926/master/custom/factories.py#L309-L314

master/custom/factories.py
```diff
@@ -1365,10 +1365,21 @@ def setup(self, **kwargs):
...
             ),
             Compile(
-                name="Compile host libFFI",
+                name="Compile host libffi",
                 command=["python3", "Tools/wasm/emscripten", "make-libffi"],
                 env=compile_environ,
             ),
+        ])
+        if branch != '3.14':
+            # Can enable on 3.14 if/when python/cpython#161845 is merged
```
StanFromIreland, line 1374:
```suggestion
            # Can enable on 3.14 if/when python/cpython#137066 is merged
```
I assume this was meant to point to the backport?

hoodmane:
@freakboy3742 does it look okay to merge now with the check that @StanFromIreland suggested?

freakboy3742:
The mpdec changes have been merged into 3.14, so we can roll back to the "non-gated version" and merge.

master/custom/factories.py
```diff
@@ -1365,10 +1365,21 @@ def setup(self, **kwargs):
...
+            # Can enable on 3.14 if/when python/cpython#137066 is merged
+            self.addStep(
+                Compile(
+                    name="Compile host libmpdec",
+                    command=["python3", "Tools/wasm/emscripten", "make-mpdec"],
+                    env=compile_environ,
+                )
+            )
+        self.addSteps([
             Configure(
```
zware, line 1383:
```suggestion
            Compile(
                name="Compile host libmpdec",
                command=["python3", "Tools/wasm/emscripten", "make-mpdec"],
                env=compile_environ,
            ),
            Configure(
```

zware (commented):
IIUC the condition is no longer necessary

PR #616, opened 2025-07-22
Emscripten: Add smoke test for using PyRepl in Chrome

hoodmane:
Depends on:
https://github.com/python/cpython/pull/137004

### DIFF
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -1396,7 +1396,6 @@ class EmscriptenBuild(BaseBuild):
             Test(
                 name="PyRepl in Chrome smoke test",
                 command=[
-                    "bash",
                     "run_test.sh",
                 ],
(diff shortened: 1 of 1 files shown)


master/custom/factories.py
```diff
@@ -1392,6 +1393,16 @@ def setup(self, **kwargs):
                 env=compile_environ,
                 timeout=step_timeout(self.test_timeout),
             ),
+            Test(
+                name="PyRepl in Chrome smoke test",
+                command=[
+                    "bash",
```
freakboy3742, line 1399: The `bash` shouldn't be required here (it doesn't hurt, but it's inconsistent).

freakboy3742:
python/cpython#137004 has now been backported to 3.14 (as python/cpython#137067), so this is good to merge.

PR #622, opened 2025-08-27
Add `BUILDBOT` envvar

StanFromIreland:
From mailing list discussion: https://mail.python.org/archives/list/python-buildbots@python.org/thread/CFBDCVYANYZCT3TYKCCEO6QGKQI2XUVD/

I slightly modified @gpshead 's naming suggestion;-)


StanFromIreland:
Whenever the buildmaster is restated, I would assume that is done after a commit?

jaraco:
I just ran some tests and unfortunately, [some buildbot runners](https://buildbot.python.org/#/builders/1677/builds/172) don't yet have the setting.

StanFromIreland:
> `test_environ` is used in setup

I did not realize, the name is not particularly clear. I will look in to it.

PR #625, opened 2025-08-29
Correct the working directory for Emscripten PyRepl tests

freakboy3742:
An update to #624 - it looks like the issue was the working directory not including the OOT `build` prefix.


freakboy3742:
@encukou I've installed the dependencies on the worker; and it looks like we've got a clean build! Thanks for the debugging work on this.

PR #633, opened 2025-10-05
Add `savannah-raspbian` worker

### LINKED ISSUE #632: New worker request: savannah-rpi5-arm64
### Username

savannah

### GitHub username

savannahostrowski

### Name

Savannah Ostrowski

### Email address

savannah@python.org

### Password status

I need a new owner password

### Processor architecture

arm64

### Operating System

Debian GNU/Linux 12 (bookworm)

### Anything special about the worker?

This is a Raspberry Pi 5 with 4 cores, 8GB RAM, active cooling (official Pi cooler), NVMe SSD storage (256GB).

savannahostrowski:
Closes #632

### DIFF
--- a/master/custom/builders.py
+++ b/master/custom/builders.py
@@ -22,6 +22,7 @@ from custom.factories import (
     ClangUbsanFunctionLinuxBuild,
     ClangUnixInstalledBuild,
     SharedUnixBuild,
+    SlowDebugUnixBuild,
     SlowNonDebugUnixBuild,
     SlowUnixInstalledBuild,
     NonDebugUnixBuild,
@@ -161,7 +162,6 @@ STABLE_BUILDERS_TIER_3 = [
     # Linux armv7l (32-bit) GCC
     ("ARM Raspbian", "gps-raspbian", SlowNonDebugUnixBuild),
     ("ARM64 Raspbian", "stan-raspbian", SlowNonDebugUnixBuild),
-    ("ARM64 Raspbian", "savannah-raspbian", SlowNonDebugUnixBuild),
 
     # FreBSD x86-64 clang
     ("AMD64 FreeBSD", "ware-freebsd", UnixBuild),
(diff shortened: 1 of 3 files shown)


master/custom/builders.py
```diff
@@ -161,6 +161,7 @@
     # Linux armv7l (32-bit) GCC
     ("ARM Raspbian", "gps-raspbian", SlowNonDebugUnixBuild),
     ("ARM64 Raspbian", "stan-raspbian", SlowNonDebugUnixBuild),
+    ("Arm64 Raspbian", "savannah-raspbian", SlowNonDebugUnixBuild),
```
savannahostrowski, line 164:
```suggestion
    ("ARM64 Raspbian", "savannah-raspbian", SlowNonDebugUnixBuild),
```

master/custom/builders.py
```diff
@@ -161,6 +161,7 @@
     # Linux armv7l (32-bit) GCC
     ("ARM Raspbian", "gps-raspbian", SlowNonDebugUnixBuild),
     ("ARM64 Raspbian", "stan-raspbian", SlowNonDebugUnixBuild),
+    ("ARM64 Raspbian", "savannah-raspbian", SlowNonDebugUnixBuild),
```
StanFromIreland, line 164: These can’t be the same, they will conflict.

```suggestion
    ("ARMv8 Raspbian", "savannah-raspbian", SlowNonDebugUnixBuild),
```

How about this, per the [spec](https://en.m.wikipedia.org/wiki/ARM_Cortex-A76)?

StanFromIreland: Or, since the Pi is more powerful, it could build debug?

itamaro (changes requested):
1. avoid duplicate builder name
2. consider initially adding this builder to the unstable builders list, and moving it to stable only after it proved stable for a few days / weeks

savannahostrowski:
Ah got it - thanks, I missed that 🤦‍♀️

PR #635, opened 2025-10-07
Replace diegorusso AArch64 worker

diegorusso:
The old machine has been decommisioned and replaced with a new more powerful one.
This is an AmpereOne with 192 cores and 500GB of memory.


StanFromIreland:
> 192 cores

That's a few cores.

diegorusso:
I moved under unstable because of the folloiwng issues:
* `test_urllib2net` will fail as the outbound ftp connection is blocked for now. I've requested to unblock it.
* `test_embed` is sensitive to the virtual environment where buildbot is installed. I've run it now without activating it but calling directly.
* `test_bigmem` used to timeout but it should be OK now.

vstinner:
Merged. Let's see how it goes.

PR #641, opened 2025-11-08
Exclude `network` resource for `{stan, pablogsal}-raspbian`

StanFromIreland:


master/custom/factories.py
```diff
@@ -112,6 +114,19 @@ def setup(self, parallel, branch, test_with_PTY=False, **kwargs):
...
         if not has_option("-j", testopts):
             testopts.append("-j2")
+        # Add excluded test resources
+        if exclude_test_resources:
+            u_loc = False
+            for i, opt in enumerate(testopts):
+                if opt.startswith("-u"):
+                    u_loc = i
+                    break
+            if u_loc:
```
pablogsal: If u_loc is the first option this won't trigger. You need to either check for false explicitly or set it to None and check for None explicitly

pablogsal (approved):

PR #642, opened 2025-11-08
Add `'urlfetch'` to excluded resources for `{stan, pablogsal}-raspbian`

StanFromIreland:
Like `'network'`, it is used for tests that pull from `http://testpython.net`, so to avoid the long times (since they need to time-out), I suggest also excluding it.


vstinner (approved):
LGTM

PR #644, opened 2025-11-16
Add a `dependabot.yml`

StanFromIreland:
Copied from [tzdata](https://github.com/python/tzdata/blob/master/.github/dependabot.yml).

Currently, both of the GH Actions we use are behind in version.


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: monthly
+    groups:
+      actions:
+        patterns:
+          - "*"
```
zware: Nit: missing trailing newline.
```suggestion
          - "*"

```

PR #652, opened 2026-01-01
Remove ARM Raspbian ASAN builder and reduce parallelism

pablogsal:
Remove ARM Raspbian ASAN builder  from the stable builders list
because ASAN builds require significantly more memory than regular
builds, making them impractical on the resource-constrained Raspberry
Pi hardware.

The parallel test count for the pablogsal-rasp worker was reduced from
2 to 1 to account for this higher memory usage. ASAN instrumentation
typically increases memory consumption by 2-10x, and running multiple
test processes in parallel on a Raspberry Pi can exhaust available
memory and cause build failures.


master/custom/builders.py
StanFromIreland: Should it not be moved here?

pablogsal: Which one? This one I am just removing

PR #695, opened 2026-03-31
Pass the worker to the build factory

encukou:
We used to pluck the following attributes from CPythonWorker, our "dataclass-ish" config object, and pass them to builds:
- parallel_tests (via lookup in `parallel`, a global dict)
- timeout_factor (via `extra_factory_args`)
- exclude_test_resources (via `extra_factory_args`)

Instead, pass in the whole CPythonWorker.

Also: Simplify if/else logic for adding the -j option, with a helper that can return an empty list.

Also: Remove "-j2" from class-level testFlags: it always matched the default or was overridden.

### DIFF
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -38,6 +38,21 @@ def step_timeout(timeout):
     return timeout + 10 * 60
 
 
+def get_j_opts(worker, default=None):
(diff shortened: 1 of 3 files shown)


zware (approved):
LGTM :)

PR #696, opened 2026-03-31
Mark PPC64LE Fedora Stable Clang as unstable

encukou:
This buildbot has been failing for a while; unfortunately it looks like the clang bug is not getting backported (upstream nor in Fedora).

@stratakis, does that sound right?

We should mark the builder unstable until clang 22 hits the repos.


vstinner (approved):
LGTM

stratakis:
> This buildbot has been failing for a while; unfortunately it looks like the clang bug is not getting backported (upstream nor in Fedora).
> 
> @stratakis, does that sound right?
> 
> We should mark the builder unstable until clang 22 hits the repos.

For the time being it seems the request for the fix is not resolved, I'll see if I can get a fix pushed in upstream and/or Fedora.

PR #699, opened 2026-04-08
Add a compatibility shim for migrating Android build tools to Platforms

freakboy3742:
Following the pattern used by the iOS build, adds a compatibility shim so that the buildbot will *run* `Platforms/Android`, but that location will work on 3.13 and 3.14 repos that store the Android build code in `Android/android.py`.

Also updates the iOS configuration to use the same general pattern for invoking the build script.

As part of this, the `CACHE_DIR` configuration has been moved *out* of the buildbot definition, and into the environment where the buildbot is executed. This means the buildbot configuration is no longer sensitive to the specific user account that is hosting the buildbot instance.

This configuration can be landed now, without the `Platforms` migration having been made - the Android folder exists in all 3.13+ repos as of right now. Once the Platforms migration occurs the symlink step will become a no-op.

### DIFF
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -1195,8 +1195,9 @@ class _IOSSimulatorBuild(UnixBuild):
             ShellCommand(
                 name="Set up compatibility symlink",
                 command=(
-                    "[ -e Platforms/Apple ] "
-                    "|| ln -s ../Apple Platforms/Apple"
+                    "if [ ! -e Platforms/Apple ]; then"
+                    " ln -s ../Apple Platforms/Apple; "
+                    "fi"
                 ),
             ),
             # Build the full iOS XCframework, including a multi-arch simulator slice.
@@ -1277,10 +1278,11 @@ class AndroidBuild(BaseBuild):
             ShellCommand(
                 name="Set up compatibility symlink",
                 command=(
-                    "[ -e Platforms/Android ] "
-                    "|| mkdir -p Platforms "
-                    "&& ln -s ../Android Platforms/Android "
-                    "&& ln -s ../Android/android.py Platforms/Android/__main__.py"
+                    "if [ ! -e Platforms/Android ]; then"
+                    " mkdir -p Platforms;"
+                    " ln -s ../Android Platforms/Android;"
+                    " ln -s ../Android/android.py Platforms/Android/__main__.py; "
+                    "fi"
                 ),
             ),
             SetPropertyFromCommand(
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -1280,7 +1280,7 @@ class AndroidBuild(BaseBuild):
                     "[ -e Platforms/Android ] "
                     "|| mkdir -p Platforms "
                     "&& ln -s ../Android Platforms/Android "
-                    "&& ln -si ../Android/android.py Platforms/Android/__main__.py"
+                    "&& ln -s ../Android/android.py Platforms/Android/__main__.py"
                 ),
             ),
             SetPropertyFromCommand(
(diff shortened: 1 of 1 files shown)


master/custom/factories.py
```diff
@@ -1273,8 +1265,20 @@ class AndroidBuild(BaseBuild):
     """
 
     def setup(self, **kwargs):
-        android_py = "Android/android.py"
+        android_py = ["python3", "Platforms/Android"]
         self.addSteps([
+            # This symlink is needed to support pre Python 3.15 builds - it makes the
+            # top level Andrdoid folder appear in the new Platforms/Android location,
```
mhsmith:
```suggestion
            # top level Android folder appear in the new Platforms/Android location,
```

master/custom/factories.py
```diff
@@ -1273,8 +1265,20 @@ class AndroidBuild(BaseBuild):
...
+            # and links `__main__.py` to the older `android.py` script. This step can
+            # be removed when 3.14 is no longer supported.
+            ShellCommand(
+                name="Set up compatibility symlink",
+                command=(
+                    "[ -e Platforms/Android ]"
+                    "|| ln -s ../Android Platforms/Android"
```
vstinner: On Python 3.13, this command fails with:

`ln: failed to create symbolic link 'Platforms/Android': No such file or directory`

You should create the `Platforms/` directory somehow. For example, add `[ -e Platforms ] || mkdir Platforms; ` prefix to your command.

master/custom/factories.py
```diff
@@ -1273,8 +1265,20 @@ class AndroidBuild(BaseBuild):
...
+                    "&& ln -si ../Android/android.py Platforms/Android/__main__.py"
```
vstinner: For readability, you should add spaces:

```suggestion
                    "[ -e Platforms/Android ] "
                    "|| ln -s ../Android Platforms/Android "
                    "&& ln -si ../Android/android.py Platforms/Android/__main__.py"
```

mhsmith: This would be much more readable as an `if ... then` statement.

master/custom/factories.py
freakboy3742: 🤦 I forgot that 3.14 has the Platforms directory; but 3.13 doesn't. Fixed.

master/custom/factories.py
freakboy3742: It might be a little more readable in *this* file, but it won't be in the logged output (since I don't think we can preserve the newlines that would make it readable. Even then, I'm not sure a single line `if (); then ; fi` is much more readable on a single line, given how prevalent `||`/`&&` syntax is in bash generally.

master/custom/factories.py
```diff
@@ -1273,8 +1268,21 @@ class AndroidBuild(BaseBuild):
...
+            # top level Android folder appear in the new Platforms/Android location,
+            # and links `__main__.py` to the older `android.py` script. This step can
+            # be removed when 3.14 is no longer supported.
+            ShellCommand(
+                name="Set up compatibility symlink",
+                command=(
+                    "[ -e Platforms/Android ] "
+                    "|| mkdir -p Platforms "
+                    "&& ln -s ../Android Platforms/Android "
+                    "&& ln -si ../Android/android.py Platforms/Android/__main__.py"
```
mhsmith: `-i` doesn't seem useful in this context.

master/custom/factories.py
mhsmith: I find it significantly more readable, especially when the expression contains both && and || and you have to try to remember which one is higher precedence.

master/custom/factories.py
freakboy3742: ... I have no idea where that came from...

master/custom/factories.py
freakboy3742: I've had a thought that let us combine the need to indent with the need for spaces for separators; thoughts?

mhsmith (approved):

vstinner (approved):
LGTM

PR #701, opened 2026-04-10
Move worker downtime configuration to workers.py

encukou:
We use `builder.canStartBuild` for scheduled downtime, which is per-*worker* configuration.
Move it out of `master.cfg` to the appropriate places:
- config in `workers.py`
- implementation details in a new module, `worker_downtime.py`

This applies `ambv-bb-win11`'s maintenance window to the PR builds, for which it was left out in GH-589.

### DIFF
--- a/master/custom/workers.py
+++ b/master/custom/workers.py
@@ -5,6 +5,8 @@
 from functools import partial
 from zoneinfo import ZoneInfo
 
+import calendar
+
 from buildbot.plugins import worker as _worker
 
 from custom.factories import MAIN_BRANCH_NAME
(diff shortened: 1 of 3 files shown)


master/master.cfg
```diff
@@ -52,6 +52,7 @@ from custom.builders import (  # noqa: E402
     STABLE,
     ONLY_MAIN_BRANCH,
 )
+from custom.worker_downtime import no_builds_between
```
zware: Looks like a phantom import?
```suggestion
```

zware (approved):
LGTM (barring one comment).

master/custom/workers.py
```diff
@@ -55,6 +59,12 @@ def __init__(
                                             notify_on_missing=emails,
                                             keepalive_interval=KEEPALIVE)
 
+# Some of Itamar's workers are reprovisioned every Wednesday at 9am PT.
+# Builds scheduled between 8am - 10am PT on Wednesdays will be delayed to
+# 10am PT.
+itamaro_downtime = no_builds_between(
+    "8:00", "10:00", day_of_week=2, tz=ZoneInfo("America/Los_Angeles")
```
vstinner: You might add `WEDNESDAY = 2` to `worker_downtime.py` for extra clarity.

master/master.cfg
vstinner: The `ZoneInfo` import can also be removed.

vstinner (approved):
LGTM

master/custom/workers.py
encukou: It's already in `calendar` :)

PR #703, opened 2026-04-16
Use a dataclass for "branches"

encukou:
We treat the main branch specially, and we use a pseudo-branch for the Pull Request buildbots.

We currently to pass `branch` around as a string, and it's not clear what for the string is in (`'main'` vs. `'3.15'`; `'PR'` vs. `'3'`).
This PR moves to a dataclass, and gives *names* for the properties and conditions. For example:

- `not branch.is_pr` instead of `if branch not in ("3",)`
- `branch.wasm_tier == 2` instead of `branchname in {"3.11", "3.12"}`
- use `branch.version_tuple` instead of an `if` ladder starting with `if branch == MAIN_BRANCH_NAME: branch = MAIN_BRANCH_VERSION`

For maintainer convenience, running `branches.py` in a terminal will dump all the branch info:

```
name:                    'main'  '3.14'  '3.13'  '3.12'  '3.11'  '3.10'    'PR'           
version_tuple:           (3, 15) (3, 14) (3, 13) (3, 12) (3, 11) (3, 10)   None           
git_ref:                 'main'  '3.14'  '3.13'  '3.12'  '3.11'  '3.10'    None           
builddir_name:           'main'  '3.14'  '3.13'  '3.12'  '3.11'  '3.10'    'pull_request' 
builder_tag:             'main'  '3.14'  '3.13'  '3.12'  '3.11'  '3.10'    'PullRequest'  
sort_key:                -9999   -14     -13     -12     -11     -10       0              
is_main:                 True    False   False   False   False   False     False          
is_pr:                   False   False   False   False   False   False     True           
gil_only:                False   False   False   True    True    True      False          
monolithic_test_asyncio: False   False   False   False   False   True      False          
wasm_tier:               2       2       2       3       3       None      2    
```

This change  makes regular builder setup very similar to PR builder setup. This PR stops short of merging them, let's do that later in a separate commit.

---

Also: add `buildersuffix` and `tags` to `BaseBuild`, instead of accessing them through `getattr`.

### DIFF
--- a/master/custom/branches.py
+++ b/master/custom/branches.py
@@ -24,7 +24,7 @@ def generate_branches():
     yield BranchInfo(
         'main',
         version_tuple=(3, 15),
-        git_ref='main',
+        git_branch='main',
         is_main=True,
         builddir_name='main',
         builder_tag='main',
@@ -38,7 +38,7 @@ def generate_branches():
     yield BranchInfo(
         'PR',
         version_tuple=None,
-        git_ref=None,
+        git_branch=None,
         is_pr=True,
         builddir_name='pull_request',
         builder_tag='PullRequest',
@@ -53,7 +53,7 @@ def _maintenance_branch(major, minor, **kwargs):
         name=version_str,
         builder_tag=version_str,
         version_tuple=version_tuple,
-        git_ref=version_str,
+        git_branch=version_str,
         builddir_name=version_str,
         sort_key=-minor,
     )
@@ -82,7 +82,7 @@ def _maintenance_branch(major, minor, **kwargs):
 class BranchInfo:
     name: str
     version_tuple: tuple[int, int] | None
-    git_ref: str | None
+    git_branch: str | None
     builddir_name: str
     builder_tag: str
 
--- a/master/custom/branches.py
+++ b/master/custom/branches.py
@@ -1,0 +1,134 @@
+"""All the info about Buildbot branches
+
+We treat the main branch specially, and we use a pseudo-branch for the
+Pull Request buildbots.
+In older branches some config needs to be nudged -- for example,
+free-threading builds only make sense in 3.13+.
+
+Complex enough to wrap up in a dataclass: BranchInfo.
+
+
+Run this as a CLI command to print the info out:
+
+    python master/custom/branches.py
+
+"""
+
+import dataclasses
+from functools import total_ordering
+from typing import Any
+
+# Buildbot configuration first; see below for the BranchInfo class.
+
+def generate_branches():
+    yield BranchInfo(
+        'main',
+        version_tuple=(3, 15),
+        git_ref='main',
+        is_main=True,
+        builddir_name='main',
+        builder_tag='main',
+        sort_key=-9999,
+    )
+    yield _maintenance_branch(3, 14)
+    yield _maintenance_branch(3, 13)
+    yield _maintenance_branch(3, 12)
+    yield _maintenance_branch(3, 11)
+    yield _maintenance_branch(3, 10)
+    yield BranchInfo(
+        'PR',
+        version_tuple=None,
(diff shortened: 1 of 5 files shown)


master/custom/factories.py
```diff
@@ -1240,7 +1242,7 @@ def setup(self, branch, *args, **kwargs):
         #
         # The symlink approach will fail for Python 3.13 *PR* builds, because
         # there's no way to identify the base branch for a PR.
-        if branch == "3.13":
+        if branch.name == "3.13":
             self.py313_setup(branch, *args, **kwargs)
```
encukou, line 1246: Branching on `name` smells, but I kept it here to match the `py313_setup` name.

master/custom/branches.py
```diff
@@ -0,0 +1,134 @@
...
+
+    return result
+
+
+@total_ordering
+@dataclasses.dataclass
+class BranchInfo:
+    name: str
+    version_tuple: tuple[int, int] | None
+    git_ref: str | None
```
vstinner: I'm a little bit confused by "git_ref" name. You may rename it to "git_branch".

master/custom/branches.py
```diff
@@ -0,0 +1,134 @@
...
+    builder_tag: str
+
+    sort_key: Any
+
+    is_main: bool = False
+    is_pr: bool = False
+
+    # Branch features.
+    # Defaults are for main (and PR), overrides are in _maintenance_branch.
+    gil_only: bool = False
```
vstinner, line 96: Nowadays, Free Threading term is more common. You may replace gil_only with its opposite value: "free_threading = True".

vstinner (commented):
This change makes the value more readable and easier to maintain, thanks! Here are some suggestions.

master/custom/branches.py
encukou: OK, I guess Git terminology isn't commonly known.
Reusing “branch” for something else is confusing too, though.

master/custom/branches.py
encukou: But, the opposite would be `both_free_threading_and_gil = True` :)
Let's use `free_threading_only = True` when the GIL builds are dropped.

vstinner: Instead of `if 'nogil' in tags and branch.gil_only: continue`, I would expect something like `if 'nogil' in tags and not branch.support_free_threading: continue`. But it's not a big deal, I'm fine with `gil_only` if you prefer to keep it.

vstinner (approved):
LGTM

master/master.cfg
```diff
@@ -257,7 +248,7 @@ for git_url, branchname, git_branch in git_branches:
         builder = util.BuilderConfig(
             name=buildername,
             workernames=[worker.name],
-            builddir="%s.%s%s" % (branchname, worker.name, getattr(f, "buildersuffix", "")),
+            builddir=f"{branch.builddir_name}.{worker.name}{f.buildersuffix}",
```
zware, line 251: Sidenote: I'd really like to get us switched to using the *builder* name rather than the *worker* name for build directories, but it's probably going to require coordination with owners since there may be a bunch of (very large, due to the size of the checkout) dead directories left sitting around after such a switch.

It occurs to me that you may be introducing a nice way to do that progressively with this PR; we can set `name_builddir_after_builder = True` on new branches and use that to choose the builddir name here.

This is all just musing for the future, though, not actually related to this PR :)

master/custom/branches.py
zware: I'm fine with `gil_only` here for now, especially since we're still using the `nogil` tag for FT.  We should probably migrate both towards 'free-threading' verbiage at some point, but I'm not in a rush for it.

zware (approved):
LGTM!

PR #707, opened 2026-04-27
Decommission ambv-bb-win11

ambv:


vstinner:
Thanks for providing the buildbot worker for many months, it was useful!

PR #708, opened 2026-05-02
Build Oddballs with `--enable-pystats`

StanFromIreland:
To catch issues like: https://github.com/python/cpython/issues/149282


diegorusso (approved):
Thanks!

PR #713, opened 2026-05-05
Move 'AMD64 FreeBSD Refleaks' builds to FreeBSD14

zware:
FreeBSD 13 is now EoL, so I'm working towards retiring my `ware-freebsd` worker.

@opsec, are you okay with this change moving the FreeBSD Refleaks builder to your FreeBSD 14 worker?

### DIFF
--- a/master/custom/builders.py
+++ b/master/custom/builders.py
@@ -169,7 +169,7 @@ STABLE_BUILDERS_TIER_3 = [
 
     # FreBSD x86-64 clang
     ("AMD64 FreeBSD", "ware-freebsd", UnixBuild),
-    ("AMD64 FreeBSD Refleaks", "ware-freebsd", UnixRefleakBuild),
+    ("AMD64 FreeBSD Refleaks", "opsec-fbsd14", UnixRefleakBuild),
     ("AMD64 FreeBSD14", "opsec-fbsd14", UnixBuild),
 
     # iOS


vstinner (commented):
Why not adding Refleaks jobs to FreeBSD 15? FreeBSD 15 is now ready for production: https://www.freebsd.org/. FreeBSD 14 is the previous version.

zware:
I'm fine with either, mostly just went with 14 due to proximity :).  We are probably due to bump 15 up to stable, though.

vstinner:
> We are probably due to bump 15 up to stable, though.

The FreeBSD15 buildbot is green for at least 3 weeks, so I'm fine with promoting it to stable: https://buildbot.python.org/#/builders/1238

opsec:
The 15bot is running a not that recent version of 15. I had plans to update it to 15.0-REL, but...
The 14bot is running 14.3p13, but an update to 14.4pX is planned.

What are Refleaks builds ?

opsec:
Ah, ref-leaks builds. Interesting...

vstinner:
Refleaks checks for memory leaks, file descriptor leaks, Python reference leaks, etc.

encukou:
or from the buildbot operator view: Refleaks are the slow builds that don't run as often.

zware:
I'm going to take lack of dissent as assent and go ahead with this :)

@opsec if this turns out to be more load than you want on your worker, we can always just drop it.

PR #716, opened 2026-05-08
Only run `diegorusso-aarch64-bigmem` on `3.x` (aka `main`)

hugovk:
Can we say `branches=['3.x']`? That'll be better than always tagging on the new branch.

cc @diegorusso

### DIFF
--- a/master/custom/workers.py
+++ b/master/custom/workers.py
@@ -170,7 +170,7 @@ def get_workers(settings):
         cpw(
             name="diegorusso-aarch64-bigmem",
             tags=['linux', 'unix', 'ubuntu', 'arm', 'arm64', 'aarch64', 'bigmem'],
-            not_branches=['3.10', '3.11', '3.12', '3.13', '3.14'],
+            branches=['3.x'],
             parallel_tests=8,
             # This worker runs pyperformance for speed.python.org at 12am UTC.
             # The pyperformance run lasts less than 2h.


diegorusso:
It's a good idea but we need to change this https://github.com/python/buildmaster-config/blob/7e8da94b68cb4777dd38025630eb71897475ec00/master/master.cfg#L208 to manage the match as well.

hugovk:
I think that's okay?

Here:

```python
        if worker.not_branches and branch.name in worker.not_branches:
            continue
```

`worker.not_branches` is `None`, so this `continue` does not run.

Then because `branch.name` is `"3.x"` here:

https://github.com/python/buildmaster-config/blob/7e8da94b68cb4777dd38025630eb71897475ec00/master/custom/branches.py#L24-L32

Then here `worker.branches == ["3.x"]` and also won't run `continue` when `"3.x" in ["3.x"]`:

```python
        if worker.branches and branch.name not in worker.branches:
            continue
```

So the build should be created.

diegorusso:
ok, cool then! I missed that!

encukou (approved):
Reconfig looks happy!
```
2026-05-08 14:04:08+0000 [-] Loading configuration from '/srv/buildbot/master/master.cfg'
2026-05-08 14:04:09+0000 [-] adding 0 new builders, removing 1
```

PR #721, opened 2026-05-15
Upgrade Twisted to 26.4.0

vstinner:


vstinner:
This PR replaces https://github.com/python/buildmaster-config/pull/718 which fix security issues.

PR #723, opened 2026-05-17
Retire ware-freebsd worker

zware:
FreeBSD 13 has reached end-of-life, and newer FreeBSDs are
already well covered.


vstinner:
Thanks for running the FreeBSD 13 worker!

PR #725, opened 2026-05-26
Add CentOS Stream 10 worker and builders

stratakis:


vstinner:
LGTM. I suppose that you reuse existing passwords. Let's see how it goes.

vstinner:
The 3.13, 3.14, 3.15 and main branches are green on the buildbot worker: https://buildbot.python.org/#/workers/157.

PR #728, opened 2026-05-26
Update itamaro workers configs

itamaro:
1. Updated not_branches to exclude Python 3.13 for all of my workers. 3.13t is no longer of interest, given 3.14t exists and is considered "supported".
2. Change the Windows worker to 1 parallel builder (and double parallel tests) (context: https://github.com/python/cpython/issues/148255#issuecomment-4534972557)

### DIFF
--- a/master/custom/workers.py
+++ b/master/custom/workers.py
@@ -322,6 +322,10 @@ def get_workers(settings):
             tags=['windows', 'win-srv-22', 'amd64', 'x86-64'],
             not_branches=['3.10', '3.11', '3.12', '3.13'],
             parallel_tests=20,
+            # Parallel MSBuild builds are "unusual", and more likely to hit obscure bugs
+            # (such as file locking issues across builds)
+            # Explicit limit parallel builders to 1 to avoid such issues
+            # https://github.com/python/cpython/issues/148255#issuecomment-4534972557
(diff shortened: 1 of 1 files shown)


master/custom/workers.py
```diff
@@ -312,29 +312,29 @@ def get_workers(settings):
...
         ),
         cpw(
             name="itamaro-win64-srv-22-aws",
             tags=['windows', 'win-srv-22', 'amd64', 'x86-64'],
-            not_branches=['3.10', '3.11', '3.12'],
-            parallel_tests=10,
-            parallel_builders=2,
+            not_branches=['3.10', '3.11', '3.12', '3.13'],
+            parallel_tests=20,
+            parallel_builders=1,
```
vstinner, line 329: Maybe just remove `parallel_builders=1,` which is the default.

vstinner (approved):
LGTM

master/custom/workers.py
itamaro: considering the context, maybe it's better to make this explicit, and add a comment explaining why
I'll add a comment

vstinner: Ah yes, it's a good idea to add a comment to explain the context of this configuration.

PR #729, opened 2026-05-26
Use partial (blobless) git clones and auto-recovery on fetch failure

stratakis:
Enable filters=["blob:none"] for partial clones reducing .git/objects from ~800MB to ~250MB per build directory.

Enable clobberOnFailure so failed fetches wipe and re-clone instead of failing the build.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -139,6 +139,11 @@ GIT_KWDS = {
     # ignoring the .gitignore rules (ex: remove also ".o" files).
     "mode": "full",
     "method": "fresh",
+    # Partial (blobless) clone: only fetch commit and tree objects,
+    # blobs are fetched on demand during checkout.
+    "filters": ["blob:none"],
+    # If fetch fails, wipe and re-clone instead of failing the build.
+    "clobberOnFailure": True,
 }
 
 c["builders"] = []


stratakis:
This will save a lot of space for *newly* created build dirs, so to see the benefits, the buildbot admins should clean up their build dirs.

Basically with this change on a worker which would have 8 builders, 4 active branches, it will save roughly 17GB of disk space, plus the pr_ directories would be smaller as well.

Plus when clobberOnFailure gets active it will reclone with the blob:none option so builders that somehow failed the fetch will get the benefit (although that would depend on fetching failure), but I think it's good to have this there nonetheless.

vstinner:
I didn't know `--filter=blob:none` option. The [git-clone manual page ](https://git-scm.com/docs/git-clone) says:

> For example, `--filter=blob:none` will filter out all blobs (file contents) until needed by Git. Also, `--filter=blob:limit=<size>` will filter out all blobs of size at least `<size>`.

vstinner (approved):
LGTM

zware (approved):
I'm definitely on board with this.  I've been less than thrilled with the disk usage lately and have wondered about switching to a shallow clone freshly pulled every build, but that would entail a 45MB download on *every* build rather than a nearly 1GB download every now and then.  Blobless (maybe even treeless?) seems like a nice middle ground, assuming everybody has a new enough `git`; thank you for beating me to figuring it out :)

stratakis:
> I'm definitely on board with this. I've been less than thrilled with the disk usage lately and have wondered about switching to a shallow clone freshly pulled every build, but that would entail a 45MB download on _every_ build rather than a nearly 1GB download every now and then. Blobless (maybe even treeless?) seems like a nice middle ground, assuming everybody has a new enough `git`; thank you for beating me to figuring it out :)

treeless should save like 100+mb per fetch, but I since it needs to fetch extra on every checkout and it will slowly accumulate anyway I don't think it's worth adding it there.

I believe the blobless one would be the best for the buildbots (but maybe in the future treeless could be explored?).

From https://github.blog/open-source/git/get-up-to-speed-with-partial-clone-and-shallow-clone/ : 

> git clone --filter=blob:none <url> creates a blobless clone. These clones download all reachable commits and trees while fetching blobs on-demand. These clones are best for developers and build environments that span multiple builds.
> 
> git clone --filter=tree:0 <url> creates a treeless clone. These clones download all reachable commits while fetching trees and blobs on-demand. These clones are best for build environments where the repository will be deleted after a single build, but you still need access to commit history.
> 
> git clone --depth=1 <url> creates a shallow clone. These clones truncate the commit history to reduce the clone size. This creates some unexpected behavior issues, limiting which Git commands are possible. These clones also put undue stress on later fetches, so they are strongly discouraged for developer use. They are helpful for some build environments where the repository will be deleted after a single build.

stratakis:
Filled 2 RFEs for buildbot that could help with the situation.

vstinner:
The two buildbot RFE:

* https://github.com/buildbot/buildbot/issues/9058
* https://github.com/buildbot/buildbot/issues/9059

vstinner:
I merged the change. Let's see how it goes.

Should we suggest to buildbot owners to remove their buildbot build directories to reduce the disk space?

stratakis:
> I merged the change. Let's see how it goes.
> 
> Should we suggest to buildbot owners to remove their buildbot build directories to reduce the disk space?

Yeap, a cleanup would work right now if the changes are deployed on master, folks who have issues with disk space can immediatelly do a cleanup. Although with time and more checkouts the disk savings might not last too long but it should be a lot better.

if https://github.com/buildbot/buildbot/issues/9058 gets implemented no actions will be needed.

vstinner:
Ok, I sent an email to the buildbot list:
https://mail.python.org/archives/list/python-buildbots@python.org/thread/WI6B25C2LPG2ZFNJ6AQ64LTTONBFNIER/

PR #735, opened 2026-06-11
Make `buildbot.sh` and `make git-update-requirements` more portable

encukou:
- `Makefile`: Avoid a custom alias for `git commit`
- `buildbot.sh`: Use the directory of the script rather than the production path

---

The second one is [9f8b1095](https://github.com/zware/buildmaster-config/commit/9f8b1095a989540d1db215d772949be1b2303962) from @zware's branch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,7 @@ git-update-requirements:
 	git pull
 	git switch -c reqs main
 	make regen-requirements
-	git ci -a -m "Run make regen-requirements"
+	git commit -a -m "Run make regen-requirements"
 
 .PHONY: help
 help : Makefile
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -87,7 +87,7 @@ git-update-requirements:
 	git pull
 	git switch -c reqs main
 	make regen-requirements
-	git ci -a -m "Run make regen-requirements"
+	git commit -a -m "Run make regen-requirements"
```
vstinner, line 90: Ooops. I'm using a git alias for "git commit" for many years, and my fingers now only know "git ci" :-D Thanks for the fix.

vstinner (approved):
LGTM

zware (approved):

PR #736, opened 2026-06-11
Refactor builder setup

encukou:
*Time for the big one!*

Builders were defined as lists of triples, which is somewhat cumbersome (especially when changing the tier/stability, which means moving the definition around).

Instead, add a new dataclass, `BuilderDef`, which holds the data.

Convert a single definition to the "new" format. Use an adapter for the others, to avoid conflicts while this PR is in flight (and perhaps later -- there's not much need to convert, until a definition needs some new feature).

Turn `stability` and `tier` into generic flags, stored in a frozenset. This way, more tags can be added later, when it makes more sense here than on the worker.
Add attributes/helpers to get the stability & tier back from the tags.

Print out a formatted list of builders when the module is run as a tool (`python -m custom.builders`).

Worker selection is moved to (two lines in) `master.cfg`; I intend to work on that in the future.

The "new style" uses `factories.xxxBuild` instead of having to import each factory class individually.

---

I'm filing PRs with individual changes from @zware's branch, to review/test/fixup/revert the changes one by one. This is based on [5f8ac35](https://github.com/zware/buildmaster-config/commit/5f8ac35).
Unlike that commit, this does not prepare for multiple worker support (and worker selection by tags). I'm not forgetting that, just not doing it here :)

### DIFF
--- a/master/custom/builders.py
+++ b/master/custom/builders.py
@@ -115,7 +115,7 @@ def get_tier_from_tags(tags):
 
 # -- Dataclass-based builders -------------------------------------------
 
-# For now, most builders are defined in "generate_builderefs" calls below,
+# For now, most builders are defined in "generate_builderdefs" calls below,
 # grouped by stability and tier.
 # Feel free to convert them to a `BuilderDef` call and move them here when
 # updating them (e.g. changing stability)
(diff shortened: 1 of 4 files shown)


master/custom/builders.py
```diff
@@ -68,13 +73,75 @@
...
+    # Tests that require the 'tzdata' and 'xpickle' resources
+    BuilderDef(
+        "aarch64 Ubuntu Oddballs",
+        factories.UnixOddballsBuild,
+        tags={STABLE, TIER_1},
+        worker_name="stan-aarch64-ubuntu",
+    ),
+]
+
+def generate_builderefs(tags, tuples):
```
zware: Nit: `s/builderefs/builderdefs/` ?

zware (approved):

PR #740, opened 2026-06-22
Increase test_timeout on cstratak-*-ppc64le

encukou:
`test_tools` has been timing out on PPC64LE CentOS9 LTO. Perhaps it's time to increase the timeout.

Note the timeout was increased for RHEL8 (i.e. older CPython branches) in #468; this increases it for the other workers.

### DIFF
--- a/master/custom/workers.py
+++ b/master/custom/workers.py
@@ -131,11 +131,13 @@ def get_workers(settings):
             name="cstratak-fedora-rawhide-ppc64le",
             tags=['linux', 'unix', 'fedora', 'ppc64le'],
             parallel_tests=10,
+            timeout_factor=2,  # Increase the timeout on this slow worker
         ),
         cpw(
             name="cstratak-fedora-stable-ppc64le",
             tags=['linux', 'unix', 'fedora', 'ppc64le'],
             parallel_tests=10,
+            timeout_factor=2,  # Increase the timeout on this slow worker
         ),
         cpw(
             name="cstratak-RHEL8-ppc64le",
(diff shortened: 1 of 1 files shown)


vstinner:
@stratakis: What do you think of increasing the timeout on the Fedora/RHEL/CentOS ppc64le workers?

vstinner:
The Fedora/RHEL/CentOS ppc64 workers use `parallel_tests=10`. Maybe this value can be decreased to make the workers more stable, rather than increasing the timeout.

@stratakis: Do you recall why you chose to use `parallel_tests=10` on the ppc64 workers?

stratakis:
> The Fedora/RHEL/CentOS ppc64 workers use `parallel_tests=10`. Maybe this value can be decreased to make the workers more stable, rather than increasing the timeout.
> 
> @stratakis: Do you recall why you chose to use `parallel_tests=10` on the ppc64 workers?

The formula I recall using at the time was testing with the same number of CPUs and above that, the whole test suite run was usually faster with 2 more tests above the baseline CPU number. I tested it now and on the c9s ppc64le buildbot which has 8 cores and 8 GB of RAM.

For the full test suite on an lto build:
-j6: 7m23s, 3419M Peak memory
-j8: 5m17s, 3514M Peak memory
-j10: 4m36s, 2793M Peak memory

The timeout is the correct fix here, test_tools possibly uses a lot of CPU at the time it runs, hence competing with others.

encukou:
OK, I'll merge this.

One question though: do the `cstratak-fedora-rawhide-ppc64le`, `cstratak-fedora-stable-ppc64le`, `cstratak-RHEL8-ppc64le`, `cstratak-CentOS9-ppc64le" workers *each* have 8 cores to themselves, or are those shared?

stratakis:
> OK, I'll merge this.
> 
> One question though: do the `cstratak-fedora-rawhide-ppc64le`, `cstratak-fedora-stable-ppc64le`, `cstratak-RHEL8-ppc64le`, `cstratak-CentOS9-ppc64le" workers _each_ have 8 cores to themselves, or are those shared?

They are shared, although I don't think pinning or isolating would bring too big of a benefit, overall the host is using 40 logical CPUs our of 64 for various VMs. It won't be CPU starved but yeah different vCPUs could be running on the same core.

PR #744, opened 2026-06-24
Allow opt-in shallow cloning

zware:
Opt-in on ware-win11-arm64; something seems to be going wrong
with the blobless clones on that worker.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -204,17 +204,26 @@ for branch in BRANCHES:
 
         source = GitHub(
             repourl=str(settings.git_url),
-            timeout=3600,
             **branch_args,
-            # "git clean -fdx": remove all files which are not tracked by Git,
-            # ignoring the .gitignore rules (ex: remove also ".o" files).
-            mode="full",
-            method="fresh",
-            # Partial (blobless) clone: only fetch commit and tree objects,
-            # blobs are fetched on demand during checkout.
-            filters=["blob:none"],
-            # If fetch fails, wipe and re-clone instead of failing the build.
-            clobberOnFailure=True,
+            **{
+                # defaults
+                **dict(
+                    timeout=3600,
+                    # "git clean -fdx": remove all files which are not tracked
+                    # by Git, ignoring the .gitignore rules (ex: remove also
+                    # ".o" files).
+                    mode="full",
+                    method="fresh",
+                    # Partial (blobless) clone: only fetch commit and tree
+                    # objects, blobs are fetched on demand during checkout.
+                    filters=["blob:none"],
+                    # If fetch fails, wipe and re-clone instead of failing the
+                    # build.
+                    clobberOnFailure=True,
+                ),
+                # worker-specific config, overriding defaults
+                **worker.git_options,
+            },
         )
 
         f = builder_def.factory(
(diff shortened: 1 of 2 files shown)


master/master.cfg
```diff
@@ -204,17 +204,26 @@ for branch in BRANCHES:
...
+                    # Partial (blobless) clone: only fetch commit and tree
+                    # objects, blobs are fetched on demand during checkout.
+                    filters=["blob:none"],
+                    # If fetch fails, wipe and re-clone instead of failing the
+                    # build.
+                    clobberOnFailure=True,
+                ),
+                # worker-specific config, overriding defaults
+                **worker.git_options,
+            },
```
vstinner, line 226: I would prefer to either only use worker Git options, or only use default Git options, but not mix the two. I suggest something like:

```py
        git_options = worker.git_options
        if git_options is None:
            git_options = dict(
                # "git clean -fdx": remove all files which are not tracked
                # by Git, ignoring the .gitignore rules (ex: remove also
                # ".o" files).
                mode="full",
                method="fresh",
                # Partial (blobless) clone: only fetch commit and tree
                # objects, blobs are fetched on demand during checkout.
                filters=["blob:none"],
            )

        source = GitHub(
            repourl=str(settings.git_url),
            # If fetch fails, wipe and re-clone instead of failing the
            # build.
            clobberOnFailure=True,
            timeout=3600,
            **branch_args,
            **git_options,
        )
```

That way, you can remove `filters=None,` from ware-win11-arm64 configuration (git options).

zware: Hmm, that's the way I was going initially, but I'd like to be able to set a default timeout here and override it per-worker (for example, timeout really ought to be closer to 5 minutes than 60 on `ware-win11-arm64`).  I can also imagine allowing per-builder overrides on top of the per-worker ones.

Looking at some of my other workers, I'm wondering if the blobless clone is living up to expectations, or if the `filters` option should be an opt-in as well. @stratakis, how is blobless working out for you in practice?

stratakis: Haven't analyzed yet, I'll see if I can provide some stats one of these days. Bear in mind though that there is a bug in buildbot that the blobless fetches will grow blobfull overtime with subsequent fetches which I fixed here: https://github.com/buildbot/buildbot/pull/9074

But it requires a new release of buildbot to pull the fix.

encukou (approved):
Happy to get this in as is & possibly clean up later :)


FWIW, here's how I'd do it -- assuming a worker admin doesn't generally care about Buildbot Git step arguments: https://github.com/python/buildmaster-config/compare/main...encukou:buildmaster-config:opt-in_shallow_cloning?expand=1

stratakis:
> Opt-in on ware-win11-arm64; something seems to be going wrong with the blobless clones on that worker.

What is the exact issue you are facing with the blobless clones?

master/master.cfg
stratakis: So blobless clones work fine on my workers, promisors are there, filters work. That is of course for the first cloning, later fetches which basically happen everyday, get all the nice blobs which then accumulate several more GB, same as before. So unless the fix lands on buildbot or somehow is pulled into master it won't do much good, it will be the same as before.

vstinner (approved):
LGTM

master/master.cfg
zware: I suspect the same is happening for me; we just need your fix released :)

PR #746, opened 2026-06-25
Run pythoninfo-build on WASI

vstinner:
Replace also "make pythoninfo" with WASI "pythoninfo-host" command.

### DIFF
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -958,6 +958,12 @@ class _Wasm32WasiPreview1Build(UnixBuild):
                 command=["python3", wasi_py, "make-build-python"],
             )
         )
+        self.addStep(
+            PythonInfo(
+                name="Display build info of the build Python",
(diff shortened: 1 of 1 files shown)


vstinner:
Two builds are running with the new buildbot configuration:

* 3.14: https://buildbot.python.org/#/builders/1745/builds/1547
* 3.13: https://buildbot.python.org/#/builders/1502/builds/2254

vstinner:
The new "system" keys are logged properly by pythoninfo in the "Display build info of the build Python" step:

```
system.boot_id: 37d4e586-62cc-46b6-919b-53703418793b
system.uptime: 2 days, 0:37:00
```

`system.machine_id` is missing. The job runs in a Fedora 44 container which can explain why it's missing.

vstinner:
I just tested on a local Fedora 44 container: the `/etc/machine-id` file exists but it's empty.

PR #750, opened 2026-07-01
Move `riscv64 Ubuntu` builder to stable

StanFromIreland:
It has been passing for quite a while: https://buildbot.python.org/#/workers/104. We recently broke it, *twice*, but that'll be [fixed shortly](https://github.com/python/cpython/pull/152819), so I'd like to make it stable to avoid any regressions. Keeping this as draft for now. CC @furkanonder 

Also dropped the '23', it's running on Ubuntu 25 now. They information isn't particularly valuable (and it's also found in `test.pythoninfo`), so I opted to drop rather than keep the maintenance burden.

### DIFF
--- a/master/custom/builders.py
+++ b/master/custom/builders.py
@@ -298,6 +298,9 @@ BUILDER_DEFS.extend(generate_builderdefs({STABLE}, [
     # Linux x86 (32-bit) GCC
     ("x86 Debian Non-Debug with X", "ware-debian-x86", NonDebugUnixBuild),
     ("x86 Debian Installed with X", "ware-debian-x86", UnixInstalledBuild),
+
+    # riscv64 GCC
+    ("riscv64 Ubuntu", "onder-riscv64", SlowUnixInstalledBuild),
 ]))
 
 
@@ -414,9 +417,6 @@ BUILDER_DEFS.extend(generate_builderdefs({UNSTABLE}, [
 
     # Solaris sparcv9
     ("SPARCv9 Oracle Solaris 11.4", "kulikjak-solaris-sparcv9", UnixBuild),
-
-    # riscv64 GCC
-    ("riscv64 Ubuntu23", "onder-riscv64", SlowUnixInstalledBuild),
 
     # Arch Usan (see stable "AMD64 Arch Linux Usan Function" above)
     ("AMD64 Arch Linux Usan", "pablogsal-arch-x86_64", ClangUbsanLinuxBuild),


vstinner:
> Also dropped the '23', it's running on Ubuntu 26 now

The https://buildbot.python.org/#/builders/1438/builds/1496 build was finished one hour ago: test.pythoninfo reports Ubuntu 25.04. Is it a bug in test.pythoninfo? Or the /etc/os-release file is outdated?

```
platform.freedesktop_os_release[ID]: ubuntu
platform.freedesktop_os_release[NAME]: Ubuntu
platform.freedesktop_os_release[VERSION]: 25.04 (Plucky Puffin)
platform.freedesktop_os_release[VERSION_CODENAME]: plucky
platform.freedesktop_os_release[VERSION_ID]: 25.04
```

vstinner:
Ah, the buildbot worker description is "Ubuntu 25.04 riscv64 - StarFive VisionFive 2 v1.3B" which is consistent with test.pythoninfo.

StanFromIreland:
It's a bug in the Stan module, not `test.pythoninfo`, I wrote down the wrong number. Sorry for the confusion.

vstinner:
Well, the recent worker history contains many failing builds (especially `test_c_stack_unwind`, that has been fixed in the meanwhile). I would suggest to wait a few days to see how it goes (if it can stay green several days in a row).

StanFromIreland:
Yes, I plan to wait a while. Although note that 3.13 has been green for a long time (as it didn't get the broken tests).

furkanonder (approved):
@StanFromIreland Thanks for upgrading the buildbot to stable. It's been running for a while now and has proven to be stable.

StanFromIreland:
It has been stable for the last few days, there were two flaky tests that passed on re-runs (run [one](https://buildbot.python.org/#/builders/1726/builds/929) and [two](https://buildbot.python.org/#/builders/1438/builds/1498)), but I recognise them from other Buildbots.

Merging this evening if there are no objections. If it causes trouble we can always revert.

vstinner:
I confirm that `riscv64 Ubuntu23` builds are green in the last 5 days in 3.14, 3.15 and main branches, since `test_c_stack_unwind` was fixed in the 3.15 and main branches. `riscv64 Ubuntu23 3.13` is green for 2 months.

vstinner:
I merged the change. I don't know what happens when a buildbot worker is renamed. We will see :-)

StanFromIreland:
New builders are created, and the old ones are left without workers it seems: https://buildbot.python.org/#/builders/1379

PR #751, opened 2026-07-02
Add mechanism for specifying configure options per branch

stratakis:


stratakis:
This is for making the possibility to enable the dtrace tests on my builders on the 3.15+ branch, but the mechanism should be generic enough to allow for any combination

encukou:
Maybe this is *too* generic? We don't really need "any combination" here; generalizing from one use case is usually premature.

Could you add the dtrace flags? Without an effect (and without tests) the PR is not too useful.

stratakis:
I've added the dtrace flag for my workers. I'll sent a different PR for a more constrained approach, to see if it makes more sense.

stratakis:
superseded by https://github.com/python/buildmaster-config/pull/754

PR #752, opened 2026-07-02
Use a (smart) set of branches for branch selection

encukou:
Add a `branches` attribute on worker, builderdef & buildfactory objects. This is a set(-like) of branches the object applies to.
The set of all branches is exposed as `branches.BRANCHES`, in a custom object with some convenience APIs.

This replaced several ways of filtering which branches apply to a given build:

- `builders.ONLY_MAIN_BRANCH` -- a set of name fragments; builds with matching names are limited to the main & PR branch
- worker `branches` & `not_branches` lists: these were usually for testing a feature introduced in 3.X, or made obsolete in 3.X. I tried several variants before settling on `only_since` & `only_until`.
- `wasm_tier` & the `wasm` tag. This PR happened to bring the WASM skipping logic to one place, with comments talking about specific versions. So I removed the `wasm_tier` attribute on BranchInfo objects.

Let me know if this seems like a good direction.

---

I made the `generate_builderdefs` function from #736 take `BuilderDef` objects in addition to tuples. Avoiding unrelated moves makes for simpler diffs.

---

I'm filing PRs with individual changes from @zware's branch, to review/test/fixup/revert the changes one by one. This is based -- very loosely this time -- on [82eb39d](https://github.com/zware/buildmaster-config/commit/82eb39d).

### DIFF
--- a/master/custom/branches.py
+++ b/master/custom/branches.py
@@ -14,6 +14,7 @@ Run this as a CLI command to print the info out:
 
 """
 
+import collections.abc
 import dataclasses
 from functools import total_ordering
 from typing import Any
@@ -64,13 +65,6 @@ def _maintenance_branch(major, minor, **kwargs):
         # need more time.
         result.monolithic_test_asyncio = True
 
-    if version_tuple < (3, 11):
-        # WASM wasn't a supported platform until 3.11.
-        result.wasm_tier = None
-    elif version_tuple < (3, 13):
-        # Tier 3 support is 3.11 & 3.12.
-        result.wasm_tier = 3
-
     if version_tuple < (3, 13):
         # Free-threaded builds are available since 3.13
         result.gil_only = True
@@ -96,7 +90,6 @@ class BranchInfo:
     # Defaults are for main (and PR), overrides are in _maintenance_branch.
     gil_only: bool = False
     monolithic_test_asyncio: bool = False
-    wasm_tier: int | None = 2
 
     def __str__(self):
         return self.name
@@ -108,6 +101,9 @@ class BranchInfo:
             return NotImplemented
         return self.sort_key == other.sort_key
 
+    def __hash__(self):
+        return hash(self.sort_key)
+
     def __lt__(self, other):
         try:
             other_key = other.sort_key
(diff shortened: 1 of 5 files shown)


master/custom/branches.py
zware, line 39: (Sidenote, not for this PR): I wonder if we could get away with pulling the canonical branch info file and generating branches from it.  Or maybe rather than pulling it on startup, use Salt to pull on a schedule and stash it in a known location, and trigger a restart on changes.

master/custom/factories.py
```diff
@@ -899,6 +901,9 @@ class Wasm32WasiCrossBuild(UnixCrossBuild):
     host = "wasm32-unknown-wasi"
     host_configure_cmd = ["../../Tools/wasm/wasi-env", "../../configure"]
 
+    # See comment in _Wasm32WasiPreview1Build.__init__
+    branches = {BRANCHES[3, 11], BRANCHES[3, 12]}
```
zware, line 926: This will blow up on us when 3.11 goes away, but CI should catch it.  I'm not sure we should change anything for that fact, but it is something for us to be aware of.

master/custom/factories.py
```diff
@@ -935,6 +940,14 @@ def __init__(self, source, *, extra_tags=[], **kwargs):
         if not self.pydebug:
             extra_tags.append("nondebug")
         self.buildersuffix += self.append_suffix
+        if self.pydebug:
+            # The debug the WASI buildbot is meant for 3.11 and 3.12 only.
```
zware: Nit:


```suggestion
            # The debug WASI buildbot is meant for 3.11 and 3.12 only.
```

master/custom/workers.py
zware, line 347: Ah, I hadn't yet realized my PR was merged; I was going to update it to something very close to what you had suggested, but hadn't made it back to it yet :)

zware (approved):
LGTM :)

master/custom/branches.py
encukou: We could, but given the current/near-future state of things I think it's better as an intentional change involving a `checkconfig`.

We'd want to combine it with a clean-up PR anyway.

master/custom/factories.py
encukou: You could also spell this `.only_since(3, 11).only_until(3, 12)`, which doesn't have this problem. But I wanted to leave another example of how to pick individual branches.

StanFromIreland (approved):
LGTM

Issue #757, opened 2026-07-09
New username and password request


StanFromIreland:
Hello!

[RISE](https://riseproject.dev/) have given me access to machines to set up some Buildbots, is it possible to get a new username, `rise` and password for it? I could put it under `stan`, but I thought it would be better organised if they were separate.

Please message me on Discord or by email, `stan@python.org`. Thank you!

vstinner:
I generated a new password and sent it to Stan via Discord.

StanFromIreland:
Thank you (once again :-)!

PR #758, opened 2026-07-09
Add a new `rise-riscv64-4` worker

StanFromIreland:
This machine will be about ~twice-thrice as fast as the current Buildbot, so I think we can try debug builds on it. If it's too slow, we might have to re-consider. Draft pending https://github.com/python/buildmaster-config/issues/757.

### DIFF
--- a/master/custom/builders.py
+++ b/master/custom/builders.py
@@ -420,6 +420,9 @@ BUILDER_DEFS.extend(generate_builderdefs({UNSTABLE}, [
 
     # Arch Usan (see stable "AMD64 Arch Linux Usan Function" above)
     ("AMD64 Arch Linux Usan", "pablogsal-arch-x86_64", ClangUbsanLinuxBuild),
+
+    # RISC-V 64-bit GCC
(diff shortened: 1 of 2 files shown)


master/custom/builders.py
```diff
@@ -420,6 +420,9 @@ def generate_builderdefs(tags, tuples):
 
     # Arch Usan (see stable "AMD64 Arch Linux Usan Function" above)
     ("AMD64 Arch Linux Usan", "pablogsal-arch-x86_64", ClangUbsanLinuxBuild),
+
+    # RISC-V 64-bit GCC
```
vstinner, line 424: It's added to "Unstable No Tier builders".

vstinner (approved):
LGTM.

I generated a password for it in https://github.com/python/buildmaster-config/issues/757.

vstinner:
The new buildbot worker is running. First build on the 3.15 branch: https://buildbot.python.org/#/builders/2227/builds/1

vstinner:
The (second) "RISC-V 64-bit Ubuntu 3.15" job completed successfully! https://buildbot.python.org/#/builders/2227/builds/2

PR #760, opened 2026-07-10
Promote AMD64 Alpine Linux NoGIL worker to Stable (No Tier)

vstinner:


vstinner:
In the last 20 days, the worker was quite stable: https://buildbot.python.org/#/builders/1819.

Well, there was a test_tkinter failure 3 days ago, but it was a "legit" failure, not an issue caused by the worker.

cc @zware

zware (approved):

StanFromIreland (approved):

vstinner:
Merged. Thanks for reviews.

PR #762, opened 2026-07-17
Move `RISC-V 64-bit Ubuntu {Clang}` to stable no tier builders

StanFromIreland:
These have been green for the last week, bar a few yellows, but those should be resolved by https://github.com/python/cpython/commit/8ca85b5bc5a888b8329915aff0ceb34de9373546. Free-threading is similar, but it does seem to have a few flaky tests. I'll wait a little bit more before merging.

### DIFF
--- a/master/custom/builders.py
+++ b/master/custom/builders.py
@@ -303,8 +303,10 @@ BUILDER_DEFS.extend(generate_builderdefs({STABLE}, [
     ("x86 Debian Non-Debug with X", "ware-debian-x86", NonDebugUnixBuild),
     ("x86 Debian Installed with X", "ware-debian-x86", UnixInstalledBuild),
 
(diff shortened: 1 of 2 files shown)


encukou:
I'd wait until all the 3.12 & 3.11 variants have at least one build. Other than that, IMO they're ready to go!

StanFromIreland:
3.11 tests are broken, so I'll exclude it as well. Waiting for the fancy API proposed in https://github.com/python/buildmaster-config/pull/752.

StanFromIreland:
I decided to just use the old `not_branches` exclusion instead.

encukou:
Sorry for the wait there.

StanFromIreland:
No worries, thanks for improving the system!

PR #764, opened 2026-07-19
Add ware-ws2025 worker with MSVC refleaks and Clang builders

zware:
- **Add ware-ws2025 worker**
- **Add MSVC refleaks and Clang builders to ware-ws2025**


vstinner (approved):
LGTM

chris-eibl (approved):
Thanks a lot ❤️

PR #765, opened 2026-07-19
Update to the latest HEAD version of buildbot

zware:
PoC showing that we *could* update to buildbot/buildbot@HEAD, but I'd rather avoid it if a release is to be along before long.  I have a question to that effect in to the buildbot Discord, but have not yet had a response.

### DIFF
--- a/requirements-3.13.txt
+++ b/requirements-3.13.txt
@@ -1,28 +1,29 @@
-alembic==1.18.4
+alembic==1.18.5
 attrs==26.1.0
-autobahn==26.6.1
+autobahn==26.6.2
 Automat==25.4.16
 blinker==1.9.0
-buildbot==4.3.0
-buildbot-console-view==4.3.0
-buildbot-grid-view==4.3.0
-buildbot-waterfall-view==4.3.0
-buildbot-worker==4.3.0
-buildbot-wsgi-dashboards==4.3.0
-buildbot-www==4.3.0
-cbor2==6.1.2
+buildbot @ git+https://github.com/buildbot/buildbot.git@e055ebda6b4800b4762e45088b23b736e046cecf#subdirectory=master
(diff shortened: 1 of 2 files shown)


zware:
This has a noticeable effect on the time it takes to generate the venv, so I hope it can be either avoided or short-lived.  It's not world-endingly bad, though.

(cc @stratakis)

vstinner:
What is the motivation for using the latest HEAD version of buildbot?

zware:
1) to pick up @stratakis's fix for broken blobless cloning
2) to pick up @hugovk's dark mode

There may be other nice things buried in the history, it's been almost a year and a half since the last release :)

zware:
We could try just patching what we need, but I think that would actually be more complex than this.

vstinner:
I'm scared by the idea of running buildbot HEAD on production. I would prefer a release.

zware:
I do agree. In the absence of a release, though, we're short on options to get a fix that we need (and an update my eyes want :)).  Maybe we could set a deadline of 1 month, and if there's no release by then we go ahead?  In that time I'll try to coordinate with @JacobCoffee about database backups (I think we have them, probably, but I don't know the particulars or what a restoration would require) to be ready in case things go wrong.

stratakis:
Yeah an update would be great for the buildbots' disk space, I think waiting 1 month is reasonable.

vstinner:
> to pick up @stratakis's fix for broken blobless cloning
> to pick up @hugovk's dark mode

What's the status of these changes? Are they merged in the main branch of buildbot?

Did someone ask for a formal buildbot release? I would really prefer to stick to releases.

zware:
Both are merged and requests have been made multiple times over the course of a few months at this point.

vstinner:
> Both are merged

Oh, nice!

> requests have been made multiple times over the course of a few months at this point.

I didn't find your requests. I created https://github.com/buildbot/buildbot/issues/9126 to request a buildbot 4.4.0 release.

PR #766, opened 2026-07-22
Coerce tags to set before comparison with tier tag set

zware:


encukou:
Thank you for the fix!

PR #767, opened 2026-07-24
Use environ to set PlatformToolset for Windows Clang builder

zware:


vstinner:
So this change is for the AMD64 Windows Server 2025 Clang 3.x builder: https://buildbot.python.org/#/builders/2261.

vstinner:
Ah, the 3.14 job ran https://buildbot.python.org/#/builders/2272/builds/5 and built successfully, but 5 tests failed (test_call test_ctypes test_idle test_importlib test_peg_generator).

PR #770, opened 2026-07-27
Fix the set of branches used on bcannon-wasi

zware:


encukou:
Thank you for the quick fix, and sorry for introducing the bug.

PR #778, opened 2026-08-12
Skip test_dtrace on RHEL/Fedora on branch 3.12 and older

vstinner:
Add UnixBuild.create_test_opts(). Override this method in RHEL8Build to add "-x test_dtrace" on branch older than 3.13.

### DIFF
--- a/master/custom/factories.py
+++ b/master/custom/factories.py
@@ -471,7 +471,6 @@ class RHEL8Build(UnixBuild):
             # possible to enable it on Fedora/RHEL. But these changes were
             # only backported up to the 3.13 branch.
             # https://github.com/python/cpython/issues/98894
-            # https://github.com/python/cpython/issues/98894
             testops.extend(('-x', 'test_dtrace'))
         return testops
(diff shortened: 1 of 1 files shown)


vstinner:
test_dtrace started to fail recently on buildbot older than 3.13. Example on PPC64LE Fedora Stable 3.12: https://buildbot.python.org/#/builders/1121/builds/1085/steps/6/logs/stdio, test_dtrace fails with a timeout (a test hangs, this issue has been fixed in the main branch).

Recent activity around dtrace:

* test_dtrace got multiple fixes and enhancements to be able to enable the test on Fedora/RHEL: https://github.com/python/cpython/issues/98894.
* `dtrace` command was installed on Fedora/RHEL buildbots.

The problem is that test_dtrace changes were only backported up to 3.13. So I propose skipping test_dtrace on 3.12 and older.

cc @stratakis

encukou (approved):
Let's do this now; I'll send a refactor PR later.

PR #781, opened 2026-08-27
Move `aarch64 Ubuntu 24.04 BigMem` builder to stable

StanFromIreland:
It has been going well for quite a while: https://buildbot.python.org/#/builders/1832 (Some tests were failing for a bit due to a test bug in https://github.com/python/cpython/issues/152190, but that has since been fixed.)

I'll let it run for a few more days, so I'm marking as draft.


diegorusso:
The builder was put in unstable since we had to replace it with a new machine: https://github.com/python/buildmaster-config/pull/635
The test behaviour was something that happened in the meantime and not the cause of the unstable status.
Independently of the cause, the builder is now stable hence it can be promoted.

In my opinion, there is no need to wait a few days, let's just do it.

StanFromIreland:
> In my opinion, there is no need to wait a few days, let's just do it.

Sounds good to me! :rocket:

diegorusso (approved):

PR #784, opened 2026-09-03
Kill only the recorded master pid when stopping

stratakis:
The SIGKILL fallback used pkill on every python process of the buildbot user. It now kills the pid in master/twistd.pid, after confirming that process is a buildbot master and before "buildbot stop" can signal it.

The SIGKILL fallback in make stop-master, added for the hang in #384, was pkill -KILL -u buildbot python, which kills every python process the buildbot user owns. It now kills only the pid in master/twistd.pid, after confirming that process is a buildbot master. The check runs before buildbot stop, which signals whatever the file names without looking, so a stale pid file is removed rather than signalled.

Found the issue while working with having a second master on the host.

Tested against a recycled pid (untouched), a hung master (killed), a bystander process (survives), a malformed pid file (ignored) and a healthy master (stops normally).

Related: #384

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,11 +3,8 @@ SYSTEM_PYTHON=python$(PYTHON_VERSION)
 VENV_DIR=./venv
 REQUIREMENTS=requirements-$(PYTHON_VERSION).txt
 PIP=$(VENV_DIR)/bin/pip
-# make stop-server kills all processes named "python"
-PKILL_NAME="python"
 BUILDBOT=$(VENV_DIR)/bin/buildbot
 VENV_CHECK=$(VENV_DIR)/lib/python$(PYTHON_VERSION)/site-packages/buildbot/master.py
-USER=buildbot
 LOGLINES=50
 
 # Setup targets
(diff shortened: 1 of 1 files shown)


vstinner:
Issue gh-785 says "the master should run under a systemd unit". If the future is to use systemd, it may be better to write a systemd service configuration, rather than this shell script to read the pid file. systemd has reliable tools to start/stop a service (without the need of a pid file).

stratakis:
> Issue [gh-785](https://github.com/python/buildmaster-config/issues/785) says "the master should run under a systemd unit". If the future is to use systemd, it may be better to write a systemd service configuration, rather than this shell script to read the pid file. systemd has reliable tools to start/stop a service (without the need of a pid file).

Indeed but this needs to be resolved on the infra side and that is if folks agree on the steps I proposed. Overall it's a longer work. This fix resolves an issue that exists now. If the migration happens, the whole make target can go away then.

Makefile
```diff
@@ -65,18 +62,33 @@ start-master: run-target
 restart-master: TARGET=restart
 restart-master: run-target
 
+# Exits successfully when master/twistd.pid names a running buildbot master.
+PID_IS_MASTER = pid=`cat master/twistd.pid 2>/dev/null`; \
```
zware: Nit: I prefer the `$()` form for shell subprocesses, even in a Makefile where it has to be `$$()`
```suggestion
PID_IS_MASTER = pid=$$(cat master/twistd.pid 2>/dev/null); \
```

Makefile
```diff
@@ -65,18 +62,33 @@ start-master: run-target
...
 ## stop-master       Stop the buildbot master
-stop-master: TARGET=stop
-stop-master: run-target
-	# issue #384: sometimes when "buildbot stop master" sends SIGINT to
+stop-master: $(VENV_CHECK)
+	# issue #384: sometimes when "buildbot stop master" sends SIGTERM to
 	# Twisted, the server goes in a broken state: it's being "shut down",
 	# but it never completes. The server stays forever in this state: it is
 	# still "running" but no longer schedules new jobs. Kill the process
 	# to make sure that it goes bad to a known state (don't run anymore).
```
zware: Nit: while we're fixing the comment :)


```suggestion
	# to make sure that it goes back to a known state (don't run anymore).
```

zware (approved):
I agree with both points; we should migrate to a `systemd` unit (regardless of whether we actually do the split process thing), and this is a solid safety measure until we've done that.

stratakis:
Applied the suggestions (force pushed, hope that's ok, these are minor).

Makefile
```diff
@@ -65,18 +62,33 @@ start-master: run-target
...
-stop-master: run-target
-	# issue #384: sometimes when "buildbot stop master" sends SIGINT to
+stop-master: $(VENV_CHECK)
+	# issue #384: sometimes when "buildbot stop master" sends SIGTERM to
 	# Twisted, the server goes in a broken state: it's being "shut down",
 	# but it never completes. The server stays forever in this state: it is
 	# still "running" but no longer schedules new jobs. Kill the process
-	# to make sure that it goes bad to a known state (don't run anymore).
-	echo "Buildbot processes (look for process name: $(PKILL_NAME))"
-	pgrep -u $(USER) $(PKILL_NAME) ||:
```
encukou, line 77: Can we leave the pgrep in as debug info, in case something goes wrong and we want to kill processes manually?

Makefile
```diff
@@ -65,18 +62,33 @@ start-master: run-target
...
-	echo "Send SIGKILL to remaining buildbot processes (if any)"
-	pkill -KILL -u $(USER) $(PKILL_NAME) ||:
+	# to make sure that it goes back to a known state (don't run anymore).
+	# Validate the pid first, since buildbot stop signals it blindly, and
+	# afterwards kill only that pid, not every python process of this user.
+	@if [ -f master/twistd.pid ] && ! { $(PID_IS_MASTER); }; then \
+		echo "Ignoring stale master/twistd.pid: $$pid is not a buildbot master"; \
```
encukou, line 82: AFAIK, the `{ $(PID_IS_MASTER); }` is a subshell. Are you sure it will set `$$pid` for the echo/kill?

stratakis: Yup, the braces it make the pid set for the current shell, tested it and it workrs.

Makefile
stratakis: Resolved.

PR #787, opened 2026-09-07
Fix some "ruff check" warnings

vstinner:
Remove unused import and variables.

### DIFF
--- a/master/custom/builders.py
+++ b/master/custom/builders.py
@@ -46,7 +46,6 @@ from custom.factories import (
     Windows64PGOTailcallBuild,
     Windows64PGONoGilBuild,
     Windows64PGONoGilTailcallBuild,
-    Windows64RefleakBuild,
     Windows64ReleaseBuild,
     MacOSArmWithBrewBuild,
     MacOSArmWithBrewNoGilBuild,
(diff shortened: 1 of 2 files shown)


hugovk (approved):
👍

We could add these Ruff checks to a CI lint as a followup to prevent new ones slipping in?

StanFromIreland (approved):

StanFromIreland:
> We could add these Ruff checks to a CI lint as a followup to prevent new ones slipping in?

I'd support adding `pre-commit` with a `ruff` hook (and possibly a few others, like trailing whitespace or zizmor).

encukou:
In my opinion, this is a textbook example of unnecessary churn.

vstinner:
I would like to fix these warnings to be able to run `ruff check` manually time to time to check for real bugs. Currently, the output is too verbose to be useful (because of the fixed issues).

StanFromIreland:
I think adding specific ignores is worse than a little churn.

vstinner:
On the main branch, `ruff check` reports 191 errors with only 38 fixable. I'm no longer sure that it's worth it to fix these ruff warnings/errors. @encukou sees that as code churn. I prefer to close my PR for now.

PR #791, opened 2026-09-14
Update RISC-V tier

StanFromIreland:
With [a recent change to PEP 11](https://github.com/python/peps/commit/98ae5f42030023756b217e590020b4c72b3f439d), they are now a tier 3 platform.

Additionally, `RISC-V 64-bit Ubuntu NoGIL` is jumping from unstable to stable tier 3, as it has been [green for a while now](https://buildbot.python.org/#/workers/160).


vstinner:
The change LGTM. I'm merged it. Thanks.

PR #792, opened 2026-09-14
Update builder tiers to match PEP 11

StanFromIreland:
This makes the tiers consistent with those specified by [PEP 11](https://peps.python.org/pep-0011/) (unfortunately some changes are not quite correct for older branches, as tiers aren't specific per branch). Also, technically it isn't a complete update as RISC-V is left to https://github.com/python/buildmaster-config/pull/791 (since that includes an additional change from unstable to stable).


master/custom/builders.py
```diff
@@ -345,6 +342,9 @@ def generate_builderdefs(tags, entries):
 
     # Emscripten
     ("WASM Emscripten", "rkm-emscripten", EmscriptenBuild),
+
+    # WASI (3.11 and 3.12 only, where WASI was tier 3)
+    ("wasm32-wasi Non-Debug", "bcannon-wasi", Wasm32WasiCrossBuild),
```
StanFromIreland, line 347: Consistent with:

https://github.com/python/buildmaster-config/blob/e78921cff2ebe00b949bfcf32ba80e889f66f020/master/custom/factories.py#L984-L987

vstinner:
Oh, I merged your RISC-V change which introduced a conflict in this change. I reviewed the RISC-V first, but it would be be better to merge this change first since it's way later. Well, too late :-)

StanFromIreland:
No worries, there would be some conflicts in either order. Fixed now. :-)

PR #793, opened 2026-09-15
Stabilize Windows Server 2025 Clang on main

zware:


vstinner (approved):
LGTM. I didn't notice that it's now possible to set stable/unstable depending on the branch, that's a nice feature.

PR #794, opened 2026-09-15
Broaden `test_dtrace` skip on RHEL/Fedora on branch 3.12 and older

hugovk:
We have a lot of RHEL/Fedora buildbots failing with `test_dtrace` for 3.10 - 3.12.

PR #778 put a `test_dtrace` skip on `RHEL8Build` and its subclasses, matching the plain Fedora Stable / Fedora Rawhide / RHEL8 / CentOS9 / CentOS10 builders, and only on x86_64/ppc64le/aarch64, since s390x doesn't use those classes.

This missed:

1. non-plain variants on Fedora/RHEL workers, that inherit directly from `UnixBuild`:
   * Refleaks: `UnixRefleakBuild``
   * LTO: `LTONonDebugUnixBuild``
   * LTO + PGO: `LTOPGONonDebugBuild``
   * Clang: `ClangUnixBuild``
   * Clang Installed: `ClangUnixInstalledBuild`

2. All plain s390x builders, defined with `UnixBuild`.

So let's move the skip from the `RHEL8Build` class to the `UnixBuild`.


vstinner (approved):
LGTM

vstinner:
I checked that all Fedora, CentOS and RHEL workers have either the `fedora` tag or the `rhel` tag.

PR #796, opened 2026-09-16
In Android tests, disable Gradle daemon

mhsmith:
This should fix the problem discussed at https://github.com/python/cpython/pull/155518#issuecomment-5259725142.


master/custom/factories.py
```diff
@@ -1320,15 +1320,17 @@ class AndroidBuild(BaseBuild):
...
+    following sections:
 
     * Install everything listed under "Prerequisites".
     * Do any OS-specific setup mentioned under "Testing".
-    * If the managed emulator appears to be running out of memory, increase
-      its RAM size as described under "Testing".
     """
 
+    # Make each run more independent
+    # (https://github.com/python/cpython/pull/155518#issuecomment-5259725142).
```
vstinner, line 1333: You might copy/paste the comment rather than adding a link.

vstinner (commented):
I don't know well Android. I don't know what is this `-Dorg.gradle.daemon=false` option. Does it disable a Gradle daemon? Maybe say it in the `test_environ` comment.

master/custom/factories.py
mhsmith: I've added some details to the comment, but I think it's worth keeping the link in case any more details are added to the issue in the future.

vstinner (approved):
The change *looks* good to me, but I'm not an Android expert.

vstinner:
Should we backport the change (you didn't add the "backport" labels)?

mhsmith:
There are no backports in this repository, are there?

vstinner:
Oh! I read quickly the PR, I saw `Platforms/Android/`, and so I expected the PR to be on Python. But that's a buildbot change! You're right, no backport is needed.

vstinner:
Merged. Thanks for the long comment, its useful to me!

PR #797, opened 2026-09-18
Add x86_64 CentOS Stream 10 workers and builders

stratakis:


master/custom/builders.py
```diff
@@ -400,13 +402,22 @@ def generate_builderdefs(tags, entries):
     ("AMD64 Fedora Rawhide LTO", "cstratak-fedora-rawhide-x86_64", LTONonDebugUnixBuild),
     ("AMD64 Fedora Rawhide LTO + PGO", "cstratak-fedora-rawhide-x86_64", LTOPGONonDebugBuild),
 
+    # CentOS Stream 10 Linux x86-64 GCC
+    ("AMD64 CentOS10", "cstratak-CentOS10-x86_64", CentOS10Build),
+    ("AMD64 CentOS10 Refleaks", "cstratak-CentOS10-x86_64", UnixRefleakBuild),
+    ("AMD64 CentOS10 LTO", "cstratak-CentOS10-x86_64", LTONonDebugUnixBuild),
+    ("AMD64 CentOS10 LTO + PGO", "cstratak-CentOS10-x86_64", LTOPGONonDebugBuild),
```
vstinner, line 409: I noticed that Fedora, CentOS, RHEL workers always use LTO. See for example "aarch64 CentOS10 3.x": 
https://buildbot.python.org/#/builders/2168/builds/806. The test output says "== Python build: debug shared LTO valgrind dtrace": notice "LTO".

I'm not sure that it's useful to duplicate have "CentOS" and "CentOS LTO", it's redundant, no?

This remarks is for all Fedora, CentOS and RHEL machines, not only these new works.

I suggest to only keep "CentOS LTO" and remove ""CentOS", so it's more explicit that LTO is used.

stratakis: They share LTO but they are different builds. aarch64 CentOS10 is CentOS10Build, the Fedora/RHEL spec configuration (--with-pydebug --enable-shared --with-valgrind --with-system-libmpdec --with-system-expat etc). aarch64 CentOS10 LTO is LTONonDebugUnixBuild, a vanilla release build whose only flag is --with-lto. 

So the base builder is the debug, RPM-like one and the LTO builder is really the non-debug one.

But yeah the LTO suffix is not a good name for both. Maybe Non-Debug would be better

I'm planning to actually go through all my workers and builders and verify disrepancies like that but it's a bigger change to cram into this PR.

vstinner: Oh ok, it's more subtle than what I expected. In that case, it makes sense to have two builders.

vstinner (approved):
LGTM

vstinner:
I checked the buildbot: it's up and running!

AMD64 CentOS10 3.x: pass successfully, build in 1min 28 sec, tests in 17 min 30 sec. https://buildbot.python.org/#/builders/2297/builds/19

AMD64 CentOS10 Refleaks 3.x: test_long_task_name_is_truncated() fails: https://buildbot.python.org/#/builders/2351/builds/8

```
FAIL: test_long_task_name_is_truncated (test.test_external_inspection.TestSelfStackTrace.test_long_task_name_is_truncated)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/var/lib/buildbot/worker/cstratak-CentOS10-x86_64/3.x.cstratak-CentOS10-x86_64.refleak/build/Lib/test/test_external_inspection.py", line 457, in test_long_task_name_is_truncated
    self.assertIn("Task-1", names)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
AssertionError: 'Task-1' not found in ['Task-4', 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx']
```

AMD64 CentOS10 FIPS Only Blake2 Builtin Hash 3.x: test_dtrace fails: https://buildbot.python.org/#/builders/2350/builds/18

```
Warning -- test_dtrace leaked temporary files (4): GMfifo1004780, GMfifo1005305, stap1gsNFh, stap5qWDET
```

